PR #3, opened 2022-11-08
Prepare action for v0 release

cormacpayne:
- Update `README.md` to reflect `v0` release
- Set `--target-port` argument in `az containerapp up` only if the default target port was determined and used
- Fix `docker build` issue where relative path for `Dockerfile` was used rather than absolute path
- Use absolute path when `dockerfilePath` argument is provided and used as a part of `docker build`


daniv-msft (approved):
Looks good to me!

PR #6, opened 2022-11-15
Fix no Dockerfile scenario

cormacpayne:
Previously the `CA_GH_ACTION_DOCKERFILE_PATH` environment variable was only being set if the relative path to a provided Dockerfile was provided via the `dockerfilePath` argument, but with a previous change to also include the `appSourcePath` as the prefix for this environment variable.

This `CA_GH_ACTION_DOCKERFILE_PATH` was then always being set with _at least_ the `appSourcePath` value, causing the action to believe that a Dockerfile was provided and jumping into the `docker build` command regardless of if a Dockerfile was provided or found.

Checking that the `dockerfilePath` argument was provided before assigning a value to `CA_GH_ACTION_DOCKERFILE_PATH` should resolve this issue.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -202,7 +202,7 @@ runs:
       if: ${{ inputs.targetPort == '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' }}
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


pauld-msft (dismissed):
:shipit:

pauld-msft (approved):

pauld-msft (approved):
:shipit:

PR #11, opened 2022-12-01
Update README and logic to better outline two scenarios targeted

cormacpayne:
Documentation and examples found in the `README.md` were updated to better reflect the two target scenarios that this action supports:
- Building an application image from source or from a provided Dockerfile
- Using a previously built and pushed image

The logic found in the `action.yml` file was also updated to better reflect these scenarios, adding additional messaging when an incorrect combination of arguments was provided, and better restricting which steps were run based on provided arguments.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -224,7 +224,7 @@ runs:
         if [[ "${{ env.CA_GH_ACTION_RUNTIME_STACK }}" == "python:"* ]]; then echo "CA_GH_ACTION_TARGET_PORT=80" >> $GITHUB_ENV; else echo "CA_GH_ACTION_TARGET_PORT=8080" >> $GITHUB_ENV; fi
 
     - name: Export target port information to environment variable for Azure CLI command
-      if: ${{ inputs.targetPort == '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' }}
+      if: ${{ env.CA_GH_ACTION_TARGET_PORT != '' }}
       shell: bash
       run: |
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -138,7 +152,7 @@ runs:
         echo "CA_GH_ACTION_IMAGE_TO_BUILD=${CA_GH_ACTION_IMAGE_TO_BUILD}" >> $GITHUB_ENV
 
     - name: Determine name of image to build if not provided
-      if: ${{ inputs.imageToBuild == '' }}
+      if: ${{ inputs.acrName != '' && inputs.imageToBuild == '' }}
```
pauld-msft, line 155: Little confused by this and the task above. So in the case that `acrName` exists and `imageToBuild` does not, wouldn't we write `CA_GH_ACTION_IMAGE_TO_BUILD` to `$GITHUB_ENV` twice? Same with below, like with `CA_GH_ACTION_IMAGE_TO_DEPLOY`. Guess I'm just trying to figure out why we need to `Export name of image to deploy to environment variable` and then `Update name of image to deploy if not provided`

pauld-msft (approved):
:shipit:

action.yml
cormacpayne: Good point, there's no reason to assign it twice -- I think this was just a method to make sure the environment variable was assigned to _at least_ some value so that it could be used later, but I can add some simple if-statements to the first checks so that the value is only ever assigned once.

pauld-msft (dismissed):

pauld-msft (approved):

PR #18, opened 2023-01-19
Update image tags to 20230118.1 MCR releases

cormacpayne:
Updates the images used in this action to the official MCR releases with the `20230118.1` tag.


pauld-msft (approved):
:shipit:

Issue #17, opened 2023-01-10
Location : Canada Central


vikingua:
When I use container-apps-deploy-action I have Location : Canada Central for container-app, Log Analytics workspace and Container Apps Environment
Can it be set somehow ?

vikingua:
Looks like Container Apps still not in production and available only in few locations, and Canada Central one of

CodyCline:
@vikingua I ran into the same problem just a few days ago! I'm not sure if location parameters are supported in this workflow. A workaround I found is using the azure cli instead and invoking the command `az containerapp up` like this: 
```yml
steps:
  - name: Deploy Azure Container app
  - run: |
    az containerapp up \
      --environment "qualified-or-existing-env-name-here" \
      --name my-app \
      --resource-group ${{ secrets.RESOURCE_GROUP }} 
      --image index.docker.io/myusername/image:latest \
      --location "West US"
```
Reference: [azure cli docs](https://learn.microsoft.com/en-us/cli/azure/containerapp?view=azure-cli-latest#az-containerapp-up)

Also, here are other qualified locations you can use for deployment which were obtained with this command:

Input: 
`az provider show -n Microsoft.App --query "resourceTypes[?resourceType=='managedEnvironments'].locations"`

Output: 
```json
[
  [
    "Canada Central",
    "West Europe",
    "North Europe",
    "East US",
    "East US 2",
    "East Asia",
    "Australia East",
    "Germany West Central",
    "Japan East",
    "UK South",
    "West US",
    "Central US",
    "North Central US",
    "South Central US",
    "Korea Central",
    "Brazil South",
    "West US 3",
    "France Central",
    "South Africa North",
    "Norway East",
    "Switzerland North"
  ]
]
```
Hope this helps!

cormacpayne:
@vikingua @CodyCline Hey folks, thanks for the comments and discussion; currently, this GitHub Action doesn't support setting the location of the Container App, but that's something I'll work on adding as an argument soon.

In the meanwhile, here's a comment from a developer on the `az containerapp` module (that this action uses) regarding the location determined for the Container App:

> When creating a new ACA environment with `up`, the CLI tries to intelligently pick a suitable location (if the user didn't provide a location with `--location`) by doing the following:
> 1. Get a list of locations the user's subscription can create environments in by calling the providers API
> 2. For each location, attempt to create an environment in that location
> 3. Continue when an environment was successfully created or error out if all locations were exhausted

[...]

PR #19, opened 2023-01-23
Miscellaneous updates based on feedback

cormacpayne:
This PR resolves the following items:

- https://github.com/Azure/container-apps-deploy-action/issues/17
  - Introduces the `location` argument to specify the region that the Container App will be deployed to
  - Added documentation for this argument to call out how to find valid locations
- https://github.com/Azure/container-apps-deploy-pipelines-task/issues/4
  - Introduces the `environmentVariables` argument to specify a list of environment variables to set on the deployed Container App
- https://github.com/Azure/container-apps-deploy-pipelines-task/issues/6
  - Handles the OS of the runner that is executing the GitHub Action to account for how the pack CLI is installed
- https://github.com/Azure/container-apps-deploy-pipelines-task/issues/3
  - Improves documentation to call out how the pack CLI is being installed by the action


action.yml
```diff
@@ -94,9 +104,19 @@ runs:
...
-      run: (curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-linux.tgz" | sudo tar -C /usr/local/bin/ --no-same-owner -xzv pack)
+      run: (curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-${{ env.RUNNER_OS }}.tgz" | sudo tar -C /usr/local/bin/ --no-same-owner -xzv pack)
+
+    - name: Install pack CLI on Windows runner
+      if: ${{ runner.os == 'Windows' }}
+      shell: bash
+      run: |
+          mkdir -p $PWD/pack && cd $PWD/pack
+          curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-windows.zip" -o "pack-windows.zip"
+          7z x pack-windows.zip > /dev/null 2>&1
```
daniv-msft, line 118: [Nothing to change] For those of you that might have the same question as I did: it looks like 7z is installed by default in runners using Windows.

README.md
```diff
@@ -1,352 +1,372 @@
...
+by this action to create runnable application images for the user when the application source code is provided and no
+additional Dockerfile is provided or found. A [builder](https://buildpacks.io/docs/concepts/components/builder/) was
+created by Oryx to take in the application source code provided to this action and produce an image that could then be
+pushed to an image registry and used within a Container App to build and run the application.
+
+A stable version of the pack CLI is installed on the GitHub runner executing the task, and depending on the base OS of
+this runner, different tools will be leverage to assist with the installation:
+- On Windows runners:
+  - `curl` will be used to pull down the `.zip` containing the `pack` executable
+  - `7z` will be used to unzip the `.zip` and place the `pack` executable in a new folder
```
william-msft, line 82: Nit: would be nice to state where `pack` is placed, in Windows. Since we let specify where `pack` is placed for Linux

action.yml
```diff
@@ -94,9 +104,19 @@ runs:
         echo "Either the 'appSourcePath' or 'imageToDeploy' argument must be provided when providing the 'acrName' argument."
         exit 1
 
-    - name: Install pack CLI
+    - name: Install pack CLI on non-Windows runner
+      if: ${{ runner.os != 'Windows' }}
       shell: bash
-      run: (curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-linux.tgz" | sudo tar -C /usr/local/bin/ --no-same-owner -xzv pack)
+      run: (curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-${{ env.RUNNER_OS }}.tgz" | sudo tar -C /usr/local/bin/ --no-same-owner -xzv pack)
```
pauld-msft, line 110: Nice, looks like the [only possible values](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context) for this follow along with the [releases](https://github.com/buildpacks/pack/releases/tag/v0.27.0) naming conventions (besides casing, and casing doesn't matter). I assume that MacOS has curl and tar and uses `/user/local/bin` for path as well? I'm pretty unfamiliar with it

pauld-msft: If the runner os as macos is not supported / can't run these commands, then maybe we want to have a check in place that ensure that the runner OS is supported?

pauld-msft: but really this is just a nit

README.md
daniv-msft: Or alternately not mentioning it for Linux? I'm not sure that's super valuable for customers. Do you see cases where they need to know that?

action.yml
cormacpayne: Good callout! [Here is the list of preinstalled software for each provided (_i.e._, not managed by the user) GitHub Action runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#preinstalled-software)

README.md
cormacpayne: @daniv-msft it may be "nice-to-know" for the customer, so I just extended this sentence to call out that the executable was placed in a "pack" folder that's in the action's working directory

action.yml
cormacpayne: >  I assume that MacOS has curl and tar and uses /usr/local/bin for path as well?

Yep that's correct -- the sample workflow I was testing in used a matrix of latest runners for each OS to make sure that the above commands worked where expected, so the `curl` and `tar` commands here work for macOS as well 👍

pauld-msft (approved):
:shipit:

daniv-msft (approved):

PR #20, opened 2023-01-30
Update to argument validation

cormacpayne:
- Ensure that `appSourcePath` and `acrName` are both provided and not provided at all
- Ensure that `imageToDeploy` is provided if `appSourcePath` and `acrName` are missing


daniv-msft (approved):
Looks good to me! Thanks

PR #22, opened 2023-02-07
Send telemetry with metadata about the action execution

cormacpayne:
This PR introduces telemetry that tracks the following:
- The scenario targeted by the user
  - _i.e._, used builder, used Dockerfile, used image
- The time (in milliseconds) it took to execute the action
- The result of the action
  - _i.e._, succeeded or failed

This PR also introduces the `disableTelemetry` argument to allow users to opt out of Microsoft collecting data about their execution of the action.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -86,10 +86,26 @@ inputs:
       'A list of environment variable(s) for the container. Space-separated values in 'key=value' format. Empty string
       to clear existing values. Prefix value with 'secretref:' to reference a secret.'
     required: false
+  disableTelemetry:
+    description: |
+      'If set to "true", no telemetry will be collected by this GitHub Action. If set to "false", or if this argument is
+      not provided, telemetry will be sent to Microsoft about the Container App build and deploy scenario targeted by
+      this GitHub Action.'
+    required: false
+    default: false
 
 runs:
   using: "composite"
   steps:
+    - name: Start timer to track action execution length, in milliseconds
+      if: ${{ always() && inputs.disableTelemetry == 'false' }}
+      shell: bash
+      run: |
+        CA_GH_ACTION_START_MILLISECONDS=$(date +%s%N | cut -b1-13)
+        echo "CA_GH_ACTION_START_MILLISECONDS=${CA_GH_ACTION_START_MILLISECONDS}" >> $GITHUB_ENV
+        CA_GH_ACTION_RESULT_ARG="--property 'result=failed'"
+        echo "CA_GH_ACTION_RESULT_ARG=${CA_GH_ACTION_RESULT_ARG}" >> $GITHUB_ENV
+
     - name: Check for ACR name provided with application source path
       if: ${{ inputs.appSourcePath != '' && inputs.acrName == '' }}
       shell: bash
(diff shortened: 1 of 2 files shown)


cormacpayne:
TODO: Waiting for new CLI image (with `oryx telemetry` and telemetry collection) to be released so I can update the images used by the builder and this workflow.

action.yml
```diff
@@ -86,10 +86,26 @@ inputs:
...
+      not provided, telemetry will be sent to Microsoft about the Container App build and deploy scenario targeted by
+      this GitHub Action.'
+    required: false
+    default: false
 
 runs:
   using: "composite"
   steps:
+    - name: Start timer to track action execution length, in milliseconds
+      if: ${{ always() && inputs.disableTelemetry == 'false' }}
```
daniv-msft, line 101: OK I'm not an expert in Github Action, but is this `always()` something we should remove from the `if`? Or is there are legitimate reason to keep it (in which case the rest of the condition doesn't seem useful?)?

daniv-msft: Similar question for the other always() below

action.yml
```diff
@@ -288,3 +318,46 @@ runs:
...
+      shell: bash
+      run: |
+        CA_GH_ACTION_SCENARIO_ARG="--property 'scenario=used-builder'"
+        echo "CA_GH_ACTION_SCENARIO_ARG=${CA_GH_ACTION_SCENARIO_ARG}" >> $GITHUB_ENV
+
+    - name: Log telemetry for action
+      if: ${{ always() && inputs.disableTelemetry == 'false' }}
+      shell: bash
+      run: |
+        docker run --rm mcr.microsoft.com/oryx/build:github-actions-debian-buster-20230206.2  /bin/bash -c "oryx telemetry --event-name 'ContainerAppsGitHubAction' ${{ env.CA_GH_ACTION_LENGTH_MILLISECONDS_ARG }} ${{ env.CA_GH_ACTION_SCENARIO_ARG }} ${{ env.CA_GH_ACTION_RESULT_ARG }}"
```
daniv-msft, line 363: Could you please clarify the impact of this line? Are we going to download and run the whole image here? If yes, any thoughts on the impact in terms of performance and all?

action.yml
cormacpayne: From the [GitHub documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs), `always()` is run independent of the preceding steps results, _i.e._, if any step fails prior to the step which includes `always()` in the conditional, the step will still run (since workflows will terminate when the action fails).

I'm using the `always()` conditional in a few places here to ensure that even if the action fails, we can still record telemetry about the run, specifically how long it executed for and that the result was `failed` (this may be able to help us catch timeouts/long-running actions, for example). I thought about adding an additional property for failed runs for which step it failed on, but decided against it since it would be a bit more involved and we can decide later if that's a piece of information we're interested in collecting.

action.yml
daniv-msft: Thanks for clarifying. Yeah that's pretty big. Let's sync during our 1:1 to see if we have any workaround.

cormacpayne: Update: after some testing, the time it takes to both pull the latest CLI image on MCR and run the `oryx telemetry` command is around ~4s, which is a more acceptable hit to performance.

action.yml
```diff
@@ -360,4 +360,4 @@ runs:
       if: ${{ always() && inputs.disableTelemetry == 'false' }}
       shell: bash
       run: |
-        docker run --rm mcr.microsoft.com/oryx/build:github-actions-debian-buster-20230206.2  /bin/bash -c "oryx telemetry --event-name 'ContainerAppsGitHubAction' ${{ env.CA_GH_ACTION_LENGTH_MILLISECONDS_ARG }} ${{ env.CA_GH_ACTION_SCENARIO_ARG }} ${{ env.CA_GH_ACTION_RESULT_ARG }}"
\ No newline at end of file
+        docker run --rm mcr.microsoft.com/oryx/cli:debian-buster-20230207.2  /bin/bash -c "oryx telemetry --event-name 'ContainerAppsGitHubAction' ${{ env.CA_GH_ACTION_LENGTH_MILLISECONDS_ARG }} ${{ env.CA_GH_ACTION_SCENARIO_ARG }} ${{ env.CA_GH_ACTION_RESULT_ARG }}"
```
daniv-msft, line 363: Out of curiosity, could you please clarify why not using the same image everywhere (for example `mcr.microsoft.com/oryx/cli:builder-debian-buster-20230208.1`)? I assumed it will get the same `oryx telemetry` capability, but maybe there is a subtlety? Or is it because this image is smaller?

cormacpayne: The CLI image we use for the builder is bigger because we pre-bake more into it since we cannot currently defer installing the platform-specific `apt-get` calls (and other build-time dependencies) when we generate the build script, so it makes it tricky to justify using that since it's not guaranteed the user is targeting the builder and will already have pulled this image down. I'm using the latest CLI image that Oryx publishes since it _does_ rely on this logic, which makes the image significantly smaller, and since we only need the image to call `oryx telemetry`, it makes the pulling and command call much quicker.

daniv-msft: Thanks for clarifying!

daniv-msft (approved):

PR #23, opened 2023-02-10
Ensure the pack CLI is only installed for the builder scenario

cormacpayne:
- Re-order the steps such that the pack CLI is only installed when the builder scenario is targeted
- Update a few conditionals to appropriately represent the scenario being targeted

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -120,20 +120,6 @@ runs:
         echo "The argument 'imageToDeploy' must be provided if neither 'appSourcePath' nor 'acrName' are provided."
         exit 1
 
-    - name: Install pack CLI on non-Windows runner
-      if: ${{ runner.os != 'Windows' }}
-      shell: bash
-      run: (curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-${{ runner.os }}.tgz" | sudo tar -C /usr/local/bin/ --no-same-owner -xzv pack)
-
-    - name: Install pack CLI on Windows runner
-      if: ${{ runner.os == 'Windows' }}
-      shell: bash
-      run: |
-          mkdir -p $PWD/pack && cd $PWD/pack
-          curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-windows.zip" -o "pack-windows.zip"
-          7z x pack-windows.zip > /dev/null 2>&1
-          rm pack-windows.zip && echo "$PWD" >> $GITHUB_PATH
-
     - name: Set Azure CLI to dynamically install missing extensions
       shell: bash
       run: az config set extension.use_dynamic_install=yes_without_prompt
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -254,7 +240,7 @@ runs:
         echo "CA_GH_ACTION_TARGET_PORT=${CA_GH_ACTION_TARGET_PORT}" >> $GITHUB_ENV
 
     - name: Determine default target port if not provided and no Dockerfile provided/found
-      if: ${{ inputs.targetPort == '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' && inputs.imageToDeploy == '' }}
+      if: ${{ inputs.targetPort == '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' && inputs.appSourcePath != '' }}
```
daniv-msft, line 243: Could you please provide more context about this change?

action.yml
```diff
@@ -280,20 +266,34 @@ runs:
...
+          rm pack-windows.zip && echo "$PWD" >> $GITHUB_PATH
+
     - name: Set Oryx++ Builder as default builder
       if: ${{ inputs.appSourcePath != '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' }}
       shell: bash
       run: pack config default-builder mcr.microsoft.com/oryx/builder:20230208.1
 
     - name: Set telemetry for Oryx++ Builder
-      if: ${{ inputs.disableTelemetry == 'false' }}
+      if: ${{ inputs.appSourcePath != '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' && inputs.disableTelemetry == 'false' }}
```
daniv-msft, line 289: I might misunderstand here but isn't the new condition going to make us send telemetry only if there is no Dockerfile? Am I missing something?

action.yml
cormacpayne: Sure thing -- I wanted to more explicit about what dictates the "builder scenario", and in our case, it's when the user provides a value for `appSourcePath` (their application source) and there's no assigned value to the environment variable `CA_GH_ACTION_DOCKERFILE_PATH` (either the `dockerfilePath` argument provided by the user or the `Dockerfile` we found in their application source).

This step is attempting to figure out the target port that the Container App should listen on if one wasn't provided, and it does so by taking the `runtimeStack` that was either provided or detected by us and choosing `80` or `8080` depending on the platform provided/detected. Since the `runtimeStack` is _only_ used in the builder case, this step should only be used when the two values mentioned above meet the conditions specified, and is independent of the value of `imageToDeploy`.

action.yml
cormacpayne: This step (and the one below) are only relevant for the builder scenario -- we either are passing in an environment variable to the builder that says "don't collect telemetry" or an environment variable that says "here's the `CALLER_ID` for the `oryx build` call we're making in the builder". Telemetry will still be collected in the Dockerfile and previously built image cases.

cormacpayne: Maybe to be a bit more clear: this step and the one below determine the value of the environment variable `CA_GH_ACTION_ORYX_BUILDER_TELEMETRY_ARG`, which is passed along to the `pack build` command call and used nowhere else, scoping just to the builder scenario.

daniv-msft: Thanks for clarifying

daniv-msft (approved):

PR #27, opened 2023-02-13
Add validation workflow to run after PR is merged to main

cormacpayne:
This PR adds a GitHub Action workflow that will run when a PR is merged into `main` and does the following:

- Builds and deploys a Container App using the following three scenarios in parallel:
  - A previously built runnable application image is provided
  - A Dockerfile is provided/found in the application source
  - A runnable application image is built from the application source using the Oryx builder
- Deletes the Container App
- Deletes the image pushed to ACR (if any)

The service principal used to perform the CRUD operations will only have Contributor access over the test resource group that we'll be using (and the ACR resource inside of it).

### DIFF
--- a/.github/workflows/run-validation-on-main.yml
+++ b/.github/workflows/run-validation-on-main.yml
@@ -1,0 +1,113 @@
+name: Run validation on main branch
+
+on:
+  push:
+    branches:
+      - main
+
+env:
+  TEST_FULL_ACR_NAME: ${{ vars.TEST_ACR_NAME }}.azurecr.io
+  TEST_IMAGE_REPOSITORY: github-actions/container-app
+
+jobs:
+  use-builder:
+    name: 'Validating builder scenario'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    env:
+      TEST_IMAGE_TAG: 'bs-${{ github.run_id }}'
+      TEST_CONTAINER_APP_NAME: 'gh-ca-bs-${{ github.run_id }}'
+
+    steps:
+      - name: Clone Oryx repository
+        uses: actions/checkout@v3
+        with:
+          repository: microsoft/Oryx
+
+      - name: Log in to Azure
+        uses: azure/login@v1
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
+
+      - name: Execute Azure Container Apps Build and Deploy Action
+        uses: Azure/container-apps-deploy-action@main
+        with:
+          appSourcePath: '${{ github.workspace }}/tests/SampleApps/DotNetCore/NetCore6PreviewWebApp'
+          acrName: ${{ vars.TEST_ACR_NAME }}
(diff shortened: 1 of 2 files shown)


.github/workflows/run-validation-on-main.yml
```diff
@@ -0,0 +1,113 @@
...
+    name: 'Validating Dockerfile scenario'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    env:
+      TEST_IMAGE_TAG: 'ds-${{ github.run_id }}'
+      TEST_CONTAINER_APP_NAME: 'gh-ca-ds-${{ github.run_id }}'
+
+    steps:
+      - name: Clone Oryx repository
```
daniv-msft, line 61: Nit: would it be possible to clone the repo/login only once, or would it be not practical?

.github/workflows/run-validation-on-main.yml
```diff
@@ -0,0 +1,113 @@
...
+        uses: Azure/container-apps-deploy-action@main
+        with:
+          appSourcePath: '${{ github.workspace }}/tests/SampleApps/DotNetCore/NetCore6PreviewWebApp'
+          acrName: ${{ vars.TEST_ACR_NAME }}
+          containerAppName: ${{ env.TEST_CONTAINER_APP_NAME }}
+          resourceGroup: ${{ vars.TEST_RESOURCE_GROUP_NAME }}
+          imageToBuild: ${{ env.TEST_FULL_ACR_NAME }}/${{ env.TEST_IMAGE_REPOSITORY }}:${{ env.TEST_IMAGE_TAG }}
+          disableTelemetry: ${{ vars.TEST_DISABLE_TELEMETRY }}
+
+      - name: Delete created Azure Container App
```
daniv-msft, line 43: I don't have a good answer for that, but if any test fail, I assume we will need some manual cleanup, right?
We could have a pre-step that tries to delete any remaining resources maybe? I don't know if there are any better solutions.

.github/workflows/run-validation-on-main.yml
```diff
@@ -0,0 +1,113 @@
+name: Run validation on main branch
```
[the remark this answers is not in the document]
cormacpayne: Discussed offline -- updated workflow so that it is triggered with any PR targeting `main`

.github/workflows/run-validation-on-main.yml
cormacpayne: Updated the workflow so the cleanup steps always run.

.github/workflows/run-validation-on-main.yml
cormacpayne: Workflow `jobs` run in separate runners, so we'll need to clone the Oryx repository onto every runner in this case.

An alternative I looked into was to use the `actions/upload-artifact` action to upload the pulled Oryx repository as an artifact of the workflow and then use the `actions/download-artifact` action on each runner to pull it down, but it felt like overkill (and probably similar performance-wise?) so decided against it.

daniv-msft (approved):
Looks good! Thanks

PR #32, opened 2023-03-01
Migrate to 'az containerapp create/update' commands

cormacpayne:
This PR focuses on the following changes:

- Migrates away from using `az containerapp up` to create/update Container Apps and instead use the `az containerapp create/update` commands to give us more control over the properties that can be passed in
- Add support for providing a YAML configuration file using the `yamlConfigPath` argument
- Add support for setting the ingress type with the `ingress` argument
- Update our validation workflow to handle more existing scenarios, as well as scenarios introduced by the YAML configuration file


.github/workflows/run-validation.yml
```diff
@@ -131,10 +214,313 @@ jobs:
...
+          WORKSPACE_NAME=$(az monitor log-analytics workspace list -g ${{ vars.TEST_RESOURCE_GROUP_NAME }} --query '[?customerId == `${{ env.CUSTOMER_ID }}`].name | [0]')
+          echo "WORKSPACE_NAME=${WORKSPACE_NAME}" >> $GITHUB_ENV
+
+      - name: Delete created Azure Container App environment
+        shell: bash
+        run: az containerapp env delete -g ${{ vars.TEST_RESOURCE_GROUP_NAME }} -n ${{ env.TEST_NEW_CONTAINER_APP_ENV }} -y
+
+      - name: Delete created workspace
+        shell: bash
+        run: az monitor log-analytics workspace delete -g ${{ vars.TEST_RESOURCE_GROUP_NAME }} -n ${{ env.WORKSPACE_NAME }} -y
```
william-msft, line 277: Should we add `if: ${{ always() }}` to both steps that perform deletes?

README.md
```diff
@@ -120,6 +123,30 @@ need to be provided in order for this action to successfully run using one of th
...
 
+### Arguments required for using a YAML configuration file
+
+| Argument name             | Required | Description |
+| ------------------------- | -------- | ----------- |
+| `yamlConfigPath`          | Yes (for this scenario) | Full path (on the executing GitHub runner) to the YAML file detailing the configuration of the Container App. |
+
+#### Important note on the YAML configuration file
+
+When providing the `yamlConfigPath`, identity arguments (_e.g._, `containerAppName`, `resourceGroup`) provided
```
daniv-msft, line 134: OK I don't know if it's just me but `identity` argument is a big ambiguous. My first thought was authentication/managed identity stuff. Maybe we could just list the 3-4 arguments that are actually impacted?

README.md
```diff
@@ -120,6 +123,30 @@ need to be provided in order for this action to successfully run using one of th
...
+to this action will still be used, but non-identity arguments (_e.g._, `containerAppEnvironment`, `environmentVariables`,
+`ingress`, etc.) will be ignored in favor of the values provided in the YAML configuration file. Identity arguments
+provided in the YAML configuration file **are not currently used** and should be provided as arguments to the action.
```
daniv-msft, line 137: I'm wondering if we should say that the logic here works exactly the same as the Container Apps CLI. I'm not saying we should remove anything from your explanation, but it might help people understand that it's not as if we reimplemented everything on our side: here this logic comes directly from the Container Apps side, and that's why we act exactly the same way.

README.md
```diff
@@ -120,6 +123,30 @@ need to be provided in order for this action to successfully run using one of th
...
+
+Image and application source arguments (_e.g._, `appSourcePath`, `imageToDeploy`) will still be used to first build
+and/or push an image that is used by the Container App; ion this case, the provided YAML configuration file will need to
```
daniv-msft, line 140: Nit: small typo, "ion"

README.md
```diff
@@ -136,6 +163,7 @@ need to be provided in order for this action to successfully run using one of th
...
 | `location`                | No       | The location that the Container App (and other created resources) will be deployed to. To view locations suitable for creating the Container App in, please run the following: `az provider show -n Microsoft.App --query "resourceTypes[?resourceType=='containerApps'].locations"` |
 | `environmentVariables`    | No       | A list of environment variable(s) for the container. Space-separated values in 'key=value' format. Empty string to clear existing values. Prefix value with 'secretref:' to reference a secret. |
+| `ingress`                 | No       | Possible options: external, internal, disabled. If set to "external" (default), the Container App will be visible from the internet or a VNET, depending on the app environment endpoint configured. If set to "internal", the Container App will be visible from within the app environment only. If set to "disabled", ingress will be disabled for this Container App and will not have an HTTP or TCP endpoint. |
```
daniv-msft, line 166: Just to be sure I get it: this one couldn't be set through the YAML and it was needed to surface it as a standalone parameter?

action.yml
```diff
@@ -113,13 +133,27 @@ runs:
         echo "The 'acrName' argument must also be provided if the 'appSourcePath' argument is provided."
         exit 1
 
-    - name: Check for application source path or previously built image when ACR name is provided
-      if: ${{ inputs.appSourcePath == '' && inputs.acrName == '' && inputs.imageToDeploy == '' }}
+    - name: Check for application source path, a previously built image or a YAML configuration file
+      if: ${{ inputs.appSourcePath == '' && inputs.imageToDeploy == '' && inputs.yamlConfigPath == '' }}
```
daniv-msft, line 137: Here I assume that we dropped the `&& inputs.acrName == ''`  check because this info might be contained in the YAML. Is that right? If yes, then should we have a slightly more complex check where we're looking for `(inputs.acrName == '' && inputs.imageToDeploy == '' && inputs.yamlConfigPath != '') || the contrary`? Is there a simplification I'm missing?

README.md
```diff
@@ -120,6 +123,30 @@ need to be provided in order for this action to successfully run using one of th
...
+### Arguments required for using a YAML configuration file
+
+| Argument name             | Required | Description |
+| ------------------------- | -------- | ----------- |
+| `yamlConfigPath`          | Yes (for this scenario) | Full path (on the executing GitHub runner) to the YAML file detailing the configuration of the Container App. |
+
+#### Important note on the YAML configuration file
+
+When providing the `yamlConfigPath`, identity arguments (_e.g._, `containerAppName`, `resourceGroup`) provided
+to this action will still be used, but non-identity arguments (_e.g._, `containerAppEnvironment`, `environmentVariables`,
```
pauld-msft, line 135: Hmm, this almost seems like it contradicts the yaml parameter explanation in https://learn.microsoft.com/en-us/cli/azure/containerapp?view=azure-cli-latest&preserve-view=true#az-containerapp-create-optional-parameters, which states that `All other parameters will be ignored.`. Is it just the required parameters that are not ignored? Specifically, the `containerAppName`, `resourceGroup` you have listed above?

README.md
pauld-msft: I believe that everything in this Additional Arguments section falls under the `non-identity` category or arguments, that is, they will be ignored in favor of whatever is configured in the [YAML](https://learn.microsoft.com/en-us/azure/container-apps/azure-resource-manager-api-spec?tabs=yaml#examples)

README.md
```diff
@@ -120,6 +123,30 @@ need to be provided in order for this action to successfully run using one of th
 | ------------------------- | -------- | ----------- |
 | `imageToDeploy`           | Yes (for this scenario) | The name of the image that has already been pushed to a registry and will be deployed to the Container App by this action. If this image is found in an ACR instance that requires authentication to pull, the `acrName` argument, or the `acrUsername` and `acrPassword` arguments, can be provided to authenticate requests to the ACR instance. |
 
+### Arguments required for using a YAML configuration file
+
+| Argument name             | Required | Description |
+| ------------------------- | -------- | ----------- |
+| `yamlConfigPath`          | Yes (for this scenario) | Full path (on the executing GitHub runner) to the YAML file detailing the configuration of the Container App. |
+
+#### Important note on the YAML configuration file
```
pauld-msft, line 132: I don't know if it is something we want to call out, or just hope that it will get fixed soon, but I do want to note that currently YAML config does not support managed identity auth.

Issue: https://github.com/microsoft/azure-container-apps/issues/524
Work item: https://msazure.visualstudio.com/One/_workitems/edit/16454602/

.github/workflows/run-validation.yml
cormacpayne: We should! Thanks for catching that -- I'll make sure all of the tear down steps have the `always()` conditional attached 👍

README.md
cormacpayne: Yes that's a good point -- let me clear that up a bit for users.

README.md
cormacpayne: Good call -- I just added a snippet at the end of this section to call out that we're using `create` and `update` and that users can refer to those documents for any additional questions that may have about the intended behavior when the YAML configuration file is provided.

README.md
cormacpayne: Both comments are correct -- ingress is a non-identifier property for a Container App. My reasoning for adding this was that the ingress concept itself felt like an important enough property on a Container App that it warranted having its own argument that users could set when using the non-YAML scenario. However, with the YAML file encapsulating this property, I'm OK with scrapping it and pointing users to the YAML scenario if they need that additional layer of configuration (which would be our current response if they were trying to update any other non-identifier property of the Container App). What do you think?

action.yml
cormacpayne: It was dropped since it's no longer needed for checking that one of our core scenarios was hit:

- `appSourcePath` -- user provides application source that they want to build OR the source that contains the Dockerfile
- `imageToDeploy` -- user provides the previously built image that they want to use for the Container App
- `yamlConfigPath` -- user provides the YAML configuration file that defines the properties of the Container App

If none of these arguments are provided, then we cannot deploy a Container App as we're missing a needed piece of information, independent of if the user provided the `acrName` argument or not. The `acrName` argument is needed when the `appSourcePath` argument is provided so we have a location to push the built image to, but if we don't have the `appSourcePath` argument, then there's no point in checking if we have the `acrName`. Hopefully that clears it up.

README.md
cormacpayne: I thought the same thing too, but the CLI behavior is a little odd -- the `--resource-group` and `--name` arguments are _required_, independent of if you've provided the `--yaml` argument.

Whatever value you provide for `--resource-group` is going to be the resource group that the Container App is created in (or that it lives in); however, whatever value you provide for `--name` is overridden by the `name` property in the YAML configuration file, so that statement they make is _almost_ true. I'll this piece of documentation to call out that only the `resourceGroup` argument for our action is necessary.

README.md
cormacpayne: Good callout 👍 I'll make sure that this is noted.

README.md
daniv-msft: I think your instinct was probably good: let's do what makes the customers the happiest. :)
Thanks both!

README.md
daniv-msft: Thanks!

daniv-msft (approved):
Looks good to me!

README.md
pauld-msft: +1 for your current approach

daniv-msft (approved):

william-msft (approved):

PR #33, opened 2023-03-09
Create resource group if it does not exist

cormacpayne:
Previously when we used the `az containerapp up` command to create a Container App, it would handle creating the resource group if it did not exist; however, when switching to the `az containerapp create` command, the expectation is that the resource group provided to the command exists, but we aren't currently handling the creation of the resource group that's provided or generated, which would cause the command to fail.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -243,6 +243,42 @@ runs:
         CA_GH_ACTION_RESOURCE_GROUP="${{ inputs.containerAppName }}-rg"
         echo "CA_GH_ACTION_RESOURCE_GROUP=${CA_GH_ACTION_RESOURCE_GROUP}" >> $GITHUB_ENV
 
+    - name: Check if the resource group exists
+      shell: bash
+      run: |
+        az group show \
+          -n ${{ env.CA_GH_ACTION_RESOURCE_GROUP }} \
+          -o none && CA_GH_ACTION_RG_EXISTS=true || CA_GH_ACTION_RG_EXISTS=false
+        echo "CA_GH_ACTION_RG_EXISTS=${CA_GH_ACTION_RG_EXISTS}" >> $GITHUB_ENV
+
+    - name: Get default location for Container Apps to use if resource group doesn't exist and location wasn't provided
+      if: ${{ env.CA_GH_ACTION_RG_EXISTS == 'false' && inputs.location == '' }}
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -243,6 +243,42 @@ runs:
...
+          -o none && CA_GH_ACTION_RG_EXISTS=true || CA_GH_ACTION_RG_EXISTS=false
+        echo "CA_GH_ACTION_RG_EXISTS=${CA_GH_ACTION_RG_EXISTS}" >> $GITHUB_ENV
+
+    - name: Get default location for Container Apps to use if resource group doesn't exist and location wasn't provided
+      if: ${{ env.CA_GH_ACTION_RG_EXISTS == 'false' && inputs.location == '' }}
+      shell: bash
+      run: |
+        CA_GH_ACTION_DEFAULT_LOCATION=$(az provider show -n Microsoft.App --query "resourceTypes[?resourceType=='containerApps'].locations[] | [0]")
+        if [[ $CA_GH_ACTION_DEFAULT_LOCATION == '' ]]; then
+          CA_GH_ACTION_DEFAULT_LOCATION="eastus2"
```
daniv-msft, line 260: Just curious: is this a default defined by Container Apps, or is it something they don't need on the CLI side?

daniv-msft (approved):

action.yml
cormacpayne: The CLI does a few things different than our flow here:
- For resource groups, they will default to [`eastus`](https://github.com/Azure/azure-cli-extensions/blob/main/src/containerapp/azext_containerapp/_utils.py#L509) if they aren't able to use the first location found in the `Microsoft.App/workspaces` RP
- For environments, they don't have a default value, but rather will try to create the environment with any of the discovered locations in the `Microsoft.App/managedEnvironments` RP, and if none are usable/found, they will throw an error

Our code is simply using the first location discovered for the `Microsoft.Apps/containerApps` RP for both resource group and environment creation, and default to `eastus2` if, for some reason, the `az provider show` returns nothing. We should probably throw an error if no location is discovered for that RP because that would imply the Container App could not later be created in that discovered location. What do you think?

cormacpayne:
Closing in favor of https://github.com/Azure/container-apps-deploy-action/pull/35

PR #35, opened 2023-03-15
Prevent existing ingress properties from being overridden

cormacpayne:
This PR aims to resolves the following issues:

- Ingress properties being overridden when updating a Container App
  - This was due to previous usage of the `az containerapp ingress enable` command, which performs a `PUT` operation on the `ingress` property and was _always_ being ran after updating an existing Container App
  - Now we will be running `az containerapp update` _only_ when we find that neither `ingress` nor `targetPort` were provided by the user, otherwise we will defer to using `az containerapp up`
- Non-existent resource groups weren't being created in the simple creation scenario


action.yml
```diff
@@ -243,6 +245,38 @@ runs:
...
+          -o none && CA_GH_ACTION_RG_EXISTS=true || CA_GH_ACTION_RG_EXISTS=false
+        echo "CA_GH_ACTION_RG_EXISTS=${CA_GH_ACTION_RG_EXISTS}" >> $GITHUB_ENV
+
+    - name: Get default location for Container Apps to use if resource group doesn't exist and location wasn't provided
+      if: ${{ env.CA_GH_ACTION_RG_EXISTS == 'false' && inputs.location == '' }}
+      shell: bash
+      run: |
+        CA_GH_ACTION_DEFAULT_LOCATION=$(az provider show -n Microsoft.App --query "resourceTypes[?resourceType=='containerApps'].locations[] | [0]")
+        CA_GH_ACTION_DEFAULT_LOCATION=$(echo $CA_GH_ACTION_DEFAULT_LOCATION | sed 's/[ !()]//g')
+        CA_GH_ACTION_DEFAULT_LOCATION=${CA_GH_ACTION_DEFAULT_LOCATION,,}
```
daniv-msft, line 262: I'm not familiar with this syntax: could you please confirm that the `,,` is on purpose?

action.yml
```diff
@@ -243,6 +245,38 @@ runs:
         CA_GH_ACTION_RESOURCE_GROUP="${{ inputs.containerAppName }}-rg"
         echo "CA_GH_ACTION_RESOURCE_GROUP=${CA_GH_ACTION_RESOURCE_GROUP}" >> $GITHUB_ENV
 
+    - name: Check if the resource group exists
```
daniv-msft, line 248: Should we only check if the resource group exists if the name of the resource group is not set as a parameter? If this parameter is set, we can assume that the customers know what they're doing (as long as the doc is clear) and avoid the extra call.

action.yml
cormacpayne: I'm glad you commented on this since I wasn't able to make a comment in the YAML file 😄 

For complete transparency about what's going on in this entire block:

- We use `az provider show` to get the locations for the `Microsoft.App/containerApps` resource type and grab the first one
- Strip out any white space and parentheses
- Make the string all lowercase letters

This allows us to take something like "East US 2" and turn it into "eastus2", normalizing the location for future command calls.

action.yml
cormacpayne: I think that's a fair assumption (as long as it's called out as you mentioned)

daniv-msft: Thanks!

daniv-msft (approved):

PR #37, opened 2023-03-20
Update registry settings, if provided, for 'up' command call

cormacpayne:
If the user provided the registry arguments (`acrName`, `acrUsername` and `acrPassword`) when deploying to an existing Container App, the corresponding `az containerapp up` command call does not include these registry values, deferring to an `az containerapp registry set` call _afterward_, which may result in the `up` command not correctly authorizing the Container App to pull the image from the registry.

This PR ensures that if these arguments were provided, they will be passed along with the `up` command call and the corresponding `registry set` call will be skipped. This does not affect the `update` command call flow where the `registry set` command call will be made afterwards.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -521,6 +521,17 @@ runs:
         CA_GH_ACTION_USE_UP="true"
         echo "CA_GH_ACTION_USE_UP=${CA_GH_ACTION_USE_UP}" >> $GITHUB_ENV
 
+    - name: Update the Container Registry details on the existing Container App
+      if: ${{ env.CA_GH_ACTION_ACR_LOGIN_ARG != '' && env.CA_GH_ACTION_USE_UP != 'true' && env.CA_GH_ACTION_RESOURCE_EXISTS == 'true' && env.CA_GH_ACTION_YAML_PROVIDED != 'true' }}
+      shell: bash
+      run: |
+        az containerapp registry set \
+          -g ${{ env.CA_GH_ACTION_RESOURCE_GROUP }} \
+          -n ${{ inputs.containerAppName }} \
+          --server ${{ inputs.acrName }}.azurecr.io \
+          --username ${{ inputs.acrUsername }} \
+          --password ${{ inputs.acrPassword }}
+
     - name: Update the existing Container App from provided arguments via 'update' (no ingress values provided)
       if: ${{ env.CA_GH_ACTION_USE_UP != 'true' && env.CA_GH_ACTION_YAML_PROVIDED != 'true' && env.CA_GH_ACTION_RESOURCE_EXISTS == 'true' }}
       shell: bash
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -525,21 +532,21 @@ runs:
...
-      if: ${{ env.CA_GH_ACTION_YAML_PROVIDED != 'true' && inputs.environmentVariables != '' }}
+      if: ${{ env.CA_GH_ACTION_USE_UP == 'true' && inputs.environmentVariables != '' }}
       shell: bash
       run: |
         CA_GH_ACTION_CONTAINER_APP_ENVIRONMENT_VARIABLES_ARG="--env-vars ${{ inputs.environmentVariables }}"
         echo "CA_GH_ACTION_CONTAINER_APP_ENVIRONMENT_VARIABLES_ARG=${CA_GH_ACTION_CONTAINER_APP_ENVIRONMENT_VARIABLES_ARG}" >> $GITHUB_ENV
 
     - name: Update the existing Container App from provided arguments via 'up' (ingress values provided)
-      if: ${{ env.CA_GH_ACTION_YAML_PROVIDED != 'true' && (inputs.targetPort != '' || inputs.ingress != '') && env.CA_GH_ACTION_RESOURCE_EXISTS == 'true' }}
+      if: ${{ env.CA_GH_ACTION_USE_UP == 'true' && env.CA_GH_ACTION_YAML_PROVIDED != 'true' && env.CA_GH_ACTION_RESOURCE_EXISTS == 'true' }}
```
harryli0108, line 549: Nit: I could be wrong here, but I'm not sure if it is necessary to check for `CA_GH_ACTION_YAML_PROVIDED` and `CA_GH_ACTION_RESOURCE_EXISTS` here since `USE_UP` can only be true if both of the prior env vars are true.

action.yml
```diff
@@ -514,8 +514,15 @@ runs:
           -n ${{ inputs.containerAppName }} \
           --yaml ${{ inputs.yamlConfigPath }}
 
+    - name: Determine whether or not 'update' or 'up' should be used
```
daniv-msft, line 517: On the long run, should this logic be done in the CLI? If yes, then should we have a work item there so that on the long run the situation is improved globally, even though in the meantime this fix will help users.

action.yml
cormacpayne: Yep, that's a good point -- we can consolidate this conditional here

action.yml
cormacpayne: I agree that, ideally, we wouldn't need to make this kind of distinction in our tooling -- the core issue is the lack of updating ingress settings through the CLI, as `az containerapp ingress enable` performs a `PUT` operation that wipes all extraneous ingress properties, and there doesn't exist an `az containerapp ingress update` command.

Unless there's another place we should open the work item, I'll open an issue in the extensions GitHub repository that requests an `ingress update` command, as well as clarifies the `ingress enable` documentation that it does not _update_ the ingress if already active.

william-msft: Yeah it would be nice if `ingress update` is part of ACA plans.

Side note:
Alternatively, we probably can leverage CLI sub-arguments in bash. However, I have found Python's to be the cleanest syntax IMO: https://stackoverflow.com/questions/10448200/how-to-parse-multiple-nested-sub-commands-using-python-argparse

william-msft (approved):

william-msft (approved):

PR #61, opened 2023-09-19
Convert composite action to typescript

snehapar9:
This change converts the existing composite action to a typescript based action so we can incorporate the private registry scenario.


src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,438 @@
...
+     * @param appSourcePath - the path to the application source on the machine
+     * @param runtimeStack - the runtime stack to use in the image layer that runs the application
+     */
+    public async createRunnableAppImage(
+        imageToDeploy: string,
+        appSourcePath: string,
+        runtimeStack: string) {
+        core.debug(`Attempting to create a runnable application image using the Oryx++ Builder with image name "${imageToDeploy}"`);
+        try {
+            let telemetryArg = `--env "CALLER_ID=azure-pipelines-rc-v1"`;
```
snehapar9, line 326: TODO: What should the `telemetryArg` value be changed to?

src/TelemetryHelper.ts
```diff
@@ -0,0 +1,107 @@
...
+                    scenarioArg = `--property 'scenario=${this.scenario}'`;
+                }
+
+                let errorMessageArg: string = '';
+                if (!util.isNullOrEmpty(this.errorMessage)) {
+                    errorMessageArg = `--property 'errorMessage=${this.errorMessage}'`;
+                }
+
+                let args: string[] = [`run`, `--rm`, `${ORYX_CLI_IMAGE}`, `/bin/bash`, `-c`, `oryx telemetry --event-name 'ContainerAppsPipelinesTaskRCV1' ` + `--processing-time '${taskLengthMilliseconds}' ${resultArg} ${scenarioArg} ${errorMessageArg}"`];
+
```
snehapar9, line 89: TODO: What should `--event-name` be?

azurecontainerapps.ts
```diff
@@ -0,0 +1,519 @@
...
+
+        // Ensure that acrName is also provided if appSourcePath is provided
+        if (!util.isNullOrEmpty(this.appSourcePath) && util.isNullOrEmpty(this.acrName)) {
+            core.error(`The 'acrName' argument must be provided when the 'appSourcePath' argument is provided.`);
+            throw Error(`The 'acrName' argument must be provided when the 'appSourcePath' argument is provided.`);
+        }
+
+        // Ensure that one of appSourcePath, imageToDeploy, or yamlConfigPath is provided
+        if (util.isNullOrEmpty(this.appSourcePath) && util.isNullOrEmpty(this.imageToDeploy) && util.isNullOrEmpty(this.yamlConfigPath)) {
+            core.error(`One of the following arguments must be provided: 'appSourcePath', 'imageToDeploy', or 'yamlConfigPath'.`);
```
cormacpayne, line 143: Looks like we're missing a `throw Error` below this line, similar to what we're doing in the above if-statement block

src/CommandHelper.ts
```diff
@@ -0,0 +1,89 @@
+import * as os from 'os'
+import * as core from '@actions/core';
+import * as exec from '@actions/exec';
```
cormacpayne, line 3: Reading more about this `@actions/exec` package, I don't think we need to differentiate between PowerShell and Bash based on the OS that the runner executing the action is hosted on. AFAIK, nothing like this was offered for Azure DevOps, so we needed to make this distinction so that we knew which shell to execute commands with, but here the `@actions/exec` library is executing CLI commands in a cross-platform manner, so it doesn't matter what the OS is or what shell is used, the library will figure that out for us.

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,438 @@
...
+                command = `Get-Content -Path ${oryxRuntimeTxtPath} -Head 1`;
+            }
+
+            const runtimeStack = await new CommandHelper().execCommandAsync(command);
+
+            // Delete the temp file
+            command = `rm ${oryxRuntimeTxtPath}`;
+            if (IS_WINDOWS_AGENT) {
+                command = `Remove-Item -Path ${oryxRuntimeTxtPath}`;
+            }
```
cormacpayne, line 384: As mentioned previously, we shouldn't need to differentiate between OS used when running commands, but we _should_ keep the distinction in this file when determining which URL to use when downloading the pack CLI (which is done later in this file)

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,438 @@
+import * as core from '@actions/core';
+import * as io from '@actions/io';
+import * as path from 'path';
+import * as os from 'os';
+import { CommandHelper } from './CommandHelper';
+import { Utility } from './Utility';
+import util = require('util');
+const cpExec = util.promisify(require('child_process').exec);
```
cormacpayne, line 8: I don't think we need to use this anymore -- we should be OK to use `@actions/exec`

src/ContainerAppHelper.ts
cormacpayne: [`github-actions-v1`](https://github.com/Azure/container-apps-deploy-action/blob/main/action.yml#L495)

src/TelemetryHelper.ts
cormacpayne: Let's update the event name to be [`ContainerAppsGitHubActionV1`](https://github.com/Azure/container-apps-deploy-action/blob/main/action.yml#L670) (I know it doesn't have the V1 suffix in the link provided, but let's make sure it points to that)

src/Utility.ts
```diff
@@ -0,0 +1,58 @@
+import * as core from '@actions/core';
+import * as exec from '@actions/exec';
+import util = require('util');
+const cpExec = util.promisify(require('child_process').exec);
```
cormacpayne, line 4: Same comment here about using `@actions/exec` instead

cormacpayne (commented):
Some initial comments to take a look at

src/Utility.ts
snehapar9: `az` commands did not execute with `@actions/exec` for some reason. Referred to [Azure/cli](https://github.com/Azure/cli/blob/master/src/main.ts#L28) and found `az` commands to execute correctly if they are executed using `util.promisify(require('child_process').exec)` instead.

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,440 @@
...
+                       `Expand-Archive`, `-LiteralPath`, `${packZipDownloadFilePath}`, `-DestinationPath`, `${PACK_CMD};`, `Remove-Item`, `-Path`, `${packZipDownloadFilePath}`];
+                commandLine = 'pwsh';
+            } else {
+                const tgzSuffix = os.platform() == 'darwin' ? 'macos' : 'linux';
+                command = `(curl -sSL \"https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-${tgzSuffix}.tgz\" | ` +
+                    'tar -C /usr/local/bin/ --no-same-owner -xzv pack)';
+                args = ['-c', command];
+                commandLine = 'bash';
+            }
+            await new Utility().executeAndthrowIfError(commandLine, args);
```
snehapar9, line 433: Even though the documentation mentions `@actions/exec` is platform agnostic, I'm not sure if the `exec` method can accept the command to be executed as the only arg and determine the `os` dynamically. The command fails by just passing the `command to be executed` as the first arg.
 From examples in [this doc](https://github.com/actions/toolkit/blob/main/packages/exec/__tests__/exec.test.ts#L69), looks like the first argument must be the command line it self like 'docker' or 'bash' and the second argument of `exec` are the args to the command line. This [example](https://github.com/actions/toolkit/blob/main/packages/exec/__tests__/exec.test.ts#L35), passes arguments to the appropriate command line after checking what the `os` is.

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
snehapar9: Updated.

src/CommandHelper.ts
snehapar9: Addressed this comment [here](https://github.com/Azure/container-apps-deploy-action/pull/61#discussion_r1335345938)

azurecontainerapps.ts
snehapar9: Updated.

azurecontainerapps.ts
```diff
@@ -0,0 +1,519 @@
...
+    /**
+     * Initializes the helpers used by this task.
+     * @param disableTelemetry - Whether or not to disable telemetry for this task.
+     */
+    private static initializeHelpers(disableTelemetry: boolean) {
+        // Set up TelemetryHelper for managing telemetry calls
+        this.telemetryHelper = new TelemetryHelper(disableTelemetry);
+
+        // Set up AzureAuthenticationHelper for managing logging in and out of Azure CLI using provided service connection
+        // this.authHelper = new AzureAuthenticationHelper();
```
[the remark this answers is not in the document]

azurecontainerapps.ts
```diff
@@ -0,0 +1,519 @@
...
+    // ACR properties
+    private static acrUsername: string;
+    private static acrPassword: string;
+
+    // Command line arguments
+    private static commandLineArgs: string[];
+
+    // Helper properties
+    private static telemetryHelper: TelemetryHelper;
+    //private static authHelper: AzureAuthenticationHelper;
```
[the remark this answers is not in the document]

azurecontainerapps.ts
```diff
@@ -0,0 +1,519 @@
...
+            this.telemetryHelper.setFailedResult(err.message);
+        } finally {
+            // If telemetry is enabled, will log metadata for this task run
+            await this.telemetryHelper.sendLogs();
+        }
+    }
+
+    // Azure DevOps build properties
+    private static buildId: string = process.env.GITHUB_RUN_ID;
+    private static buildNumber = process.env.GITHUB_RUN_NUMBER;
```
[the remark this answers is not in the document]

package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "container-apps-deploy-action",
+  "version": "1.0.0",
+  "description": "This action allows users to easily deploy their application source to an\r [Azure Container App](https://azure.microsoft.com/en-us/services/container-apps/) in their GitHub workflow by either\r providing a previously built image, a Dockerfile that an image can be built from, or using a builder to create a\r runnable application image for the user.",
```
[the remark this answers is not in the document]
snehapar9: Updated.

src/TelemetryHelper.ts
snehapar9: Thanks. Updated.

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerRegistryHelper.ts
```diff
@@ -0,0 +1,56 @@
...
+     * Authorizes Docker to make calls to the provided ACR instance using an access token that is generated via
+     * the 'az acr login --expose-token' command.
+     * @param acrName - the name of the ACR instance to authenticate calls to.
+     */
+    public async loginAcrWithAccessTokenAsync(acrName: string) {
+        core.debug(`Attempting to log in to ACR instance "${acrName}" with access token`);
+        try {
+            const command: string = `CA_ADO_TASK_ACR_ACCESS_TOKEN=$(az acr login --name ${acrName} --output json --expose-token --only-show-errors | jq -r '.accessToken'); docker login ${acrName}.azurecr.io -u 00000000-0000-0000-0000-000000000000 -p $CA_ADO_TASK_ACR_ACCESS_TOKEN > /dev/null 2>&1`;
+            const shell = os.platform() === 'win32' ? 'pwsh' : 'bash';
+            await exec.exec(shell, ['-c', command]);
```
snehapar9, line 34: Same as [this reason](https://github.com/Azure/container-apps-deploy-action/pull/61#discussion_r1335345938)

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,440 @@
...
+    /**
+     * Determines if the provided resource group exists.
+     * @param resourceGroup - the name of the resource group
+     * @returns true if the resource group exists, false otherwise
+     */
+    public async doesResourceGroupExist(resourceGroup: string): Promise<boolean> {
+        core.debug(`Attempting to determine if resource group "${resourceGroup}" exists`);
+        try {
+            const command = `group show -n ${resourceGroup} -o none`;
+            const exitCode = await exec.exec(`az`, command.split(' '));
```
snehapar9, line 198: Calling `exec` instead of `executeAndthrowIfError` to make sure pipeline does not fail and still is able to complete with a warning.

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,440 @@
...
+     * @param containerAppEnvironment - the name of the Container App Environment
+     * @param resourceGroup - the resource group that the Container App Environment is found in
+     * @returns true if the Container App Environment exists, false otherwise
+     */
+    public async doesContainerAppEnvironmentExist(containerAppEnvironment: string, resourceGroup: string): Promise<boolean> {
+        core.debug(`Attempting to determine if Container App Environment with name "${containerAppEnvironment}" exists in resource group "${resourceGroup}"`);
+        try {
+            const command = `containerapp env show -n ${containerAppEnvironment} -g ${resourceGroup} -o none`;
+            const exitCode = await exec.exec(`az`, command.split(' '));
+            return exitCode === 0;
```
snehapar9, line 182: Call `exec` instead of `executeAndthrowIfError` to make sure the pipeline does not fail and is able to complete successfully with the warning logged below.

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,440 @@
...
+     * Determines if the provided Container App exists in the provided resource group.
+     * @param containerAppName - the name of the Container App
+     * @param resourceGroup - the resource group that the Container App is found in
+     * @returns true if the Container App exists, false otherwise
+     */
+    public async doesContainerAppExist(containerAppName: string, resourceGroup: string): Promise<boolean> {
+        core.debug(`Attempting to determine if Container App with name "${containerAppName}" exists in resource group "${resourceGroup}"`);
+        try {
+            let command = `containerapp show -n ${containerAppName} -g ${resourceGroup} -o none`;
+            const exitCode = await exec.exec(`az`, command.split(' '));
```
snehapar9, line 163: Call exec instead of executeAndthrowIfError to make sure the pipeline does not fail and is able to complete successfully with the warning logged below.

azurecontainerapps.ts
```diff
@@ -0,0 +1,516 @@
+import * as core from '@actions/core';
```
cormacpayne, line 1: From what we chatted about offline: it would be great if we could refactor the code in such a way that we could easily migrate code between the GitHub Action and the Azure DevOps Task since they'll both be written in TypeScript. To do this, we need to consolidate all of the GitHub Action specific code (_i.e._, code that leverages GitHub Action packages on npm or depends on GitHub Action values) to its own helper class for this tool, and we'll have an equivalent in the Azure DevOps Task down the road. This allows the streamlined flow to be copied between the two tools, and any changes we make for one tool doesn't have to be hand-written in the other to account for how it was implemented. This ensures that we have consistent functionality across both tools and don't need to worry about tracking what is and isn't supported.

From scanning over this PR, it looks like the following components are GitHub Actions specific and would need to be ported over into a new `GitHubActionsToolHelper` file:

```
@actions/core
=============
core.getInput
core.setFailed

core.debug
core.error
core.info
core.warning


@actions/io
===========
io.which


@actions/exec
=============
exec.exec
exec.ExecOptions (I think this is used with the above method call)


Miscellaneous
=============
process.env.GITHUB_RUN_ID
process.env.GITHUB_RUN_NUMBER (we just need equivalent environment variables in ADO for these two)

getContainerAppName()
  - includes GitHub Action in naming schema (just need a function for generating the default ACA name)

Telemetry arguments
  - CALLER_ID=github-actions-v1
  - --event-name ContainerAppsGitHubActionV1 (just need a function for fetching telemetry values)
```

snehapar9: These changes have been made.

src/TelemetryHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+                if (!util.isNullOrEmpty(this.scenario)) {
+                    scenarioArg = `--property scenario=${this.scenario}`;
+                }
+
+                let errorMessageArg: string = '';
+                if (!util.isNullOrEmpty(this.errorMessage)) {
+                    errorMessageArg = `--property errorMessage=${this.errorMessage}`;
+                }
+
+                let args: string[] = [`run`, `--rm`, `${ORYX_CLI_IMAGE}`, `/bin/bash`, `-c`, `oryx telemetry --event-name ContainerAppsGitHubActionV1 ` + `--processing-time ${taskLengthMilliseconds} ${resultArg} ${scenarioArg} ${errorMessageArg}`];
```
cormacpayne, line 88: A function for fetching this `ContainerAppsGitHubActionV1` argument should be created in the tool helper class

src/TelemetryHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+const executeDockerCommand = async (args: string[], continueOnError: boolean = false): Promise<void> => {
+    try {
+        let dockerTool: string = await githubActionsToolHelper.which("docker", true);
+        await util.executeAndThrowIfError(dockerTool, args, continueOnError);
+    }
+    catch (err) {
+        githubActionsToolHelper.error(`Error: ${err.message}`);
+        throw err; // Re-throw the error
+    }
+}
```
cormacpayne, line 106: Any reason this was moved to its own function?

src/ContainerRegistryHelper.ts
```diff
@@ -0,0 +1,56 @@
+import * as os from 'os';
+import { Utility } from './Utility';
+import { GithubActionsToolHelper } from './GithubActionsToolHelper';
+
+const githubActionsToolHelper = new GithubActionsToolHelper();
```
cormacpayne, line 5: Same comment about renaming this variable

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,435 @@
...
+import * as os from 'os';
+import { Utility } from './Utility';
+import { GithubActionsToolHelper } from './GithubActionsToolHelper';
+import fs = require('fs');
+
+const ORYX_CLI_IMAGE: string = 'mcr.microsoft.com/oryx/cli:builder-debian-buster-20230208.1';
+const ORYX_BUILDER_IMAGE: string = 'mcr.microsoft.com/oryx/builder:20230208.1';
+const IS_WINDOWS_AGENT: boolean = os.platform() == 'win32';
+const PACK_CMD: string = IS_WINDOWS_AGENT ? path.join(os.tmpdir(), 'pack') : 'pack';
+const githubActionsToolHelper = new GithubActionsToolHelper();
```
cormacpayne, line 11: Same comment about renaming this variable

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,435 @@
...
+     * Gets the default location for the Container App provider.
+     * @returns the default location if found, otherwise 'eastus2'
+     */
+    public async getDefaultContainerAppLocation(): Promise<string> {
+        githubActionsToolHelper.debug(`Attempting to get the default location for the Container App service for the subscription.`);
+        try {
+            let args = [`provider`, `show`, `-n`, `Microsoft.App`, `--query`, `resourceTypes[?resourceType=='containerApps'].locations[] | [0]`];
+            let executionResult = await util.executeAndThrowIfError(`az`, args);
+            // If successful, strip out double quotes, spaces and parentheses from the first location returned
+            return !executionResult.stderr ? executionResult.stdout.toLowerCase().replace(/["() ]/g, "").trim() : `eastus2`;
```
cormacpayne, line 213: Any reason why we check the error stream here rather than the exit code like we did for previous functions?

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    /**
+     * Gets the default location for the Container App provider.
+     * @returns the default location if found, otherwise 'eastus2'
+     */
+    public async getDefaultContainerAppLocation(): Promise<string> {
+        githubActionsToolHelper.debug(`Attempting to get the default location for the Container App service for the subscription.`);
+        try {
+            let args = [`provider`, `show`, `-n`, `Microsoft.App`, `--query`, `resourceTypes[?resourceType=='containerApps'].locations[] | [0]`];
```
cormacpayne, line 210: I think we need to re-evaluate how we're constructing the command arguments that we provide to our execute function -- in this file I'm seeing some arguments that include piped values (`|`) and arguments that should be explicitly wrapped in double quotes, so I think we should chat again about the differences between what the command constructed from ADO's `tl.execSync` vs. the `core.exec` we use for GitHub Actions.

cormacpayne (changes requested):

azurecontainerapps.ts
```diff
@@ -0,0 +1,513 @@
...
+    public static async runMain(): Promise<void> {
+        let disableTelemetry = githubActionsToolHelper.getInput('disableTelemetry').toLowerCase() === 'true';
+        this.initializeHelpers(disableTelemetry);
+
+        try {
+            // Validate that the arguments provided can be used for one of the supported scenarios
+            this.validateSupportedScenarioArguments();
+
+            // Set up the Azure CLI to be used for this task
+            await this.setupAzureCli();
```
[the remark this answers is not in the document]
snehapar9: The @actions/exec methods are asynchronous in nature. When we use methods like exec.exec from the @actions/exec library in GitHub Actions workflows, they execute external commands asynchronously. 

These methods make a call to the `@actions/exec` method to execute commands so they need to be awaited here to work properly.
I'm not sure if `Azure Devops` offers a way to execute commands asynchronously, but we should probably switch if it does?

src/ContainerAppHelper.ts
snehapar9: No reason specifically, didn't think it would make a difference if we check the error stream or the exit code.
But looking at the Azure Devops task code, I think it makes sense to switch to exit code check just to be consistent.

src/GithubActionsToolHelper.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    public warning(message: string): void {
+        core.warning(message);
+    }
+
+    public debug(message: string): void {
+        core.debug(message);
+    }
+
+    public async exec(commandLine: string, args?: string[], execOptions?: exec.ExecOptions): Promise<number> {
```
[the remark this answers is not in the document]
snehapar9: In that case, this method should return a Promise object similar to [this](https://github.com/Azure/container-apps-deploy-action/blob/snehapar/container-apps-deploy-action-javascript/src/Utility.ts#L13) to make sure we can access the error code, stdout and stderr. This is a customization we had to do to make sure the output and error steams can be accessed as well because the `exec` method only returns the error code.

azurecontainerapps.ts
```diff
@@ -0,0 +1,513 @@
...
+import { GithubActionsToolHelper } from './src/GithubActionsToolHelper';
+
+const util = new Utility();
+const githubActionsToolHelper = new GithubActionsToolHelper();
+
+export class azurecontainerapps {
+
+    public static async runMain(): Promise<void> {
+        let disableTelemetry = githubActionsToolHelper.getInput('disableTelemetry').toLowerCase() === 'true';
+        this.initializeHelpers(disableTelemetry);
```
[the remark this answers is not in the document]
snehapar9: Shouldn't `disableTelemetry` be defined first? The initialization of  `TelemetryHelper` and `ContainerAppHelper` is dependant on this.

azurecontainerapps.ts
```diff
@@ -0,0 +1,513 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import { ContainerAppHelper } from './src/ContainerAppHelper';
+import { ContainerRegistryHelper } from './src/ContainerRegistryHelper';
+import { TelemetryHelper } from './src/TelemetryHelper';
+import { Utility } from './src/Utility';
+import { GithubActionsToolHelper } from './src/GithubActionsToolHelper';
+
+const util = new Utility();
+const githubActionsToolHelper = new GithubActionsToolHelper();
```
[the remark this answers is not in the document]
snehapar9: Need to clarify [this](https://github.com/Azure/container-apps-deploy-action/pull/61#discussion_r1339252620) before making this change.

azurecontainerapps.ts
```diff
@@ -0,0 +1,513 @@
...
+            this.telemetryHelper.setFailedResult(err.message);
+        } finally {
+            // If telemetry is enabled, will log metadata for this task run
+            await this.telemetryHelper.sendLogs();
+        }
+    }
+
+    // GitHub Action properties
+    private static githubRunId: string = githubActionsToolHelper.getGithubRunId();
+    private static githubRunNumber: string = githubActionsToolHelper.getGithubRunNumber();
```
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

azurecontainerapps.ts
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

src/Utility.ts
```diff
@@ -0,0 +1,55 @@
+// Note: This file is used to define utility functions that can be used across the project.
+import { GithubActionsToolHelper } from './GithubActionsToolHelper';
+
+const githubActionsToolHelper = new GithubActionsToolHelper();
```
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

azurecontainerapps.ts
```diff
@@ -0,0 +1,513 @@
...
+        // Get the previously built image to deploy, if provided
+        this.imageToDeploy = githubActionsToolHelper.getInput('imageToDeploy', { required: false });
+
+        // Get the YAML configuration file, if provided
+        this.yamlConfigPath = githubActionsToolHelper.getInput('yamlConfigPath', { required: false });
+
+        // Ensure that acrName is also provided if appSourcePath is provided
+        if (!util.isNullOrEmpty(this.appSourcePath) && util.isNullOrEmpty(this.acrName)) {
+            githubActionsToolHelper.error(`The 'acrName' argument must be provided when the 'appSourcePath' argument is provided.`);
+            throw Error(`The 'acrName' argument must be provided when the 'appSourcePath' argument is provided.`);
```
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

azurecontainerapps.ts
```diff
@@ -0,0 +1,513 @@
...
+        }
+
+        // Ensure that one of appSourcePath, imageToDeploy, or yamlConfigPath is provided
+        if (util.isNullOrEmpty(this.appSourcePath) && util.isNullOrEmpty(this.imageToDeploy) && util.isNullOrEmpty(this.yamlConfigPath)) {
+            githubActionsToolHelper.error(`One of the following arguments must be provided: 'appSourcePath', 'imageToDeploy', or 'yamlConfigPath'.`);
+            throw Error(`One of the following arguments must be provided: 'appSourcePath', 'imageToDeploy', or 'yamlConfigPath'.`);
```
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

azurecontainerapps.ts
```diff
@@ -0,0 +1,513 @@
...
+
+        // If the Container App doesn't exist, get/create the Container App Environment to use for the Container App
+        if (!this.containerAppExists) {
+            this.containerAppEnvironment = await this.getOrCreateContainerAppEnvironment(this.containerAppName, this.resourceGroup, this.location);
+        }
+    }
+
+    /**
+     * Gets the name of the Container App to use for the task. If the 'containerAppName' argument is not provided,
+     * then a default name will be generated in the form 'gh-action-app-<githubRunId>-<githubRunNumber>'.
```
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

src/GithubActionsToolHelper.ts
```diff
@@ -0,0 +1,75 @@
+import * as core from '@actions/core';
+import * as io from '@actions/io';
+import * as exec from '@actions/exec';
+
+export class GithubActionsToolHelper {
+
+    public getGithubRunId(): string {
```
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

src/GithubActionsToolHelper.ts
```diff
@@ -0,0 +1,75 @@
+import * as core from '@actions/core';
+import * as io from '@actions/io';
+import * as exec from '@actions/exec';
+
+export class GithubActionsToolHelper {
```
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

src/GithubActionsToolHelper.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    public getGithubRunId(): string {
+        return process.env['GITHUB_RUN_ID'] || '';
+    }
+
+    public getGithubRunNumber(): string {
```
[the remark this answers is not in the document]
snehapar9: :white_check_mark:

src/GithubActionsToolHelper.ts
```diff
@@ -0,0 +1,75 @@
...
+        core.error(message);
+    }
+
+    public warning(message: string): void {
+        core.warning(message);
+    }
+
+    public debug(message: string): void {
+        core.debug(message);
+    }
```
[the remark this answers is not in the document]
snehapar9: Updated.

src/GithubActionsToolHelper.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    public setFailed(message: string): void {
+        core.setFailed(message);
+    }
+
+    public which(tool: string, check?: boolean): Promise<string> {
+        return io.which(tool, check);
+    }
+
+    public getContainerAppName(containerAppName: string): string {
```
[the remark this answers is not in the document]
snehapar9: Updated

src/Utility.ts
```diff
@@ -0,0 +1,55 @@
...
+   * @param commandLine - the command to execute
+   * @param args - the arguments to pass to the command
+   * @param continueOnError - whether or not to continue execution if the command fails
+   */
+
+  public async executeAndThrowIfError(commandLine: string, args: string[], continueOnError: boolean = false): Promise<{ exitCode: number, stdout: string, stderr: string }> {
+    try {
+      let options = await githubActionsToolHelper.ExecOptions();
+      let stderr = options.listeners.stderr.toString();
+      let stdout = options.listeners.stdout.toString();
```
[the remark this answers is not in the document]
snehapar9: Updated.

src/Utility.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      let exitCode = await githubActionsToolHelper.exec(commandLine, args, options);
+
+      if (!continueOnError && exitCode !== 0) {
```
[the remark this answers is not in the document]
snehapar9: Added this so the pipeline does not fail if exit code != 0 so we are doing the same as [this](https://github.com/Azure/container-apps-deploy-pipelines-task/blob/main/azurecontainerapps/src/TelemetryHelper.ts#L91)

src/ContainerAppHelper.ts
snehapar9: Updated.

src/TelemetryHelper.ts
snehapar9: Updated.

src/TelemetryHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const SUCCESSFUL_RESULT: string = "succeeded";
+const FAILED_RESULT: string = "failed";
+
+const BUILDER_SCENARIO: string = "used-builder";
+const DOCKERFILE_SCENARIO: string = "used-dockerfile";
+const IMAGE_SCENARIO: string = "used-image";
+
+const util = new Utility();
+const githubActionsToolHelper = new GithubActionsToolHelper();
```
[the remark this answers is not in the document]
snehapar9: Updated.

src/TelemetryHelper.ts
snehapar9: Removed it.

src/Utility.ts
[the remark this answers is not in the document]
snehapar9: TODO : Make sure pipeline does not fail if telemetry fails.

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,434 @@
...
+     */
+    public async createContainerApp(
+        containerAppName: string,
+        resourceGroup: string,
+        environment: string,
+        imageToDeploy: string,
+        optionalCmdArgs: string[]) {
+        toolHelper.writeDebug(`Attempting to create Container App with name "${containerAppName}" in resource group "${resourceGroup}" based from image "${imageToDeploy}"`);
+        try {
+            let command = `az containerapp create -n ${containerAppName} -g ${resourceGroup} -i ${imageToDeploy} --environment ${environment}`;
```
cormacpayne, line 37: We need to add `--output none` to the end of this command call per the latest changes made in `v1`

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,434 @@
...
+     * @param resourceGroup - the resource group that the Container App is found in
+     * @param yamlConfigPath - the path to the YAML configuration file that the Container App properties will be based from
+     */
+    public async createContainerAppFromYaml(
+        containerAppName: string,
+        resourceGroup: string,
+        yamlConfigPath: string) {
+        toolHelper.writeDebug(`Attempting to create Container App with name "${containerAppName}" in resource group "${resourceGroup}" from provided YAML "${yamlConfigPath}"`);
+        try {
+            let command = `az containerapp create -n ${containerAppName} -g ${resourceGroup} --yaml ${yamlConfigPath}`;
```
cormacpayne, line 60: We need to add `--output none` to the end of this command call per the latest changes made in `v1`

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,434 @@
...
+     * @param optionalCmdArgs - a set of optional command line arguments
+     */
+    public async updateContainerApp(
+        containerAppName: string,
+        resourceGroup: string,
+        imageToDeploy: string,
+        optionalCmdArgs: string[]) {
+        toolHelper.writeDebug(`Attempting to update Container App with name "${containerAppName}" in resource group "${resourceGroup}" based from image "${imageToDeploy}"`);
+        try {
+            let command = `az containerapp update -n ${containerAppName} -g ${resourceGroup} -i ${imageToDeploy}`;
```
cormacpayne, line 82: We need to add `--output none` to the end of this command call per the latest changes made in `v1`

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,434 @@
...
+     * @param resourceGroup - the resource group that the existing Container App is found in
+     * @param yamlConfigPath - the path to the YAML configuration file that the Container App properties will be based from
+     */
+    public async updateContainerAppFromYaml(
+        containerAppName: string,
+        resourceGroup: string,
+        yamlConfigPath: string) {
+        toolHelper.writeDebug(`Attempting to update Container App with name "${containerAppName}" in resource group "${resourceGroup}" from provided YAML "${yamlConfigPath}"`);
+        try {
+            let command = `az containerapp update -n ${containerAppName} -g ${resourceGroup} --yaml ${yamlConfigPath}`;
```
cormacpayne, line 142: We need to add `--output none` to the end of this command call per the latest changes made in `v1`

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,434 @@
...
+
+    /**
+     * Gets the name of an existing Container App Environment in the provided resource group.
+     * @param resourceGroup - the resource group to check for an existing Container App Environment
+     * @returns the name of the existing Container App Environment, null if none exists
+     */
+    public async getExistingContainerAppEnvironment(resourceGroup: string) {
+        toolHelper.writeDebug(`Attempting to get the existing Container App Environment in resource group "${resourceGroup}"`);
+        try {
+            let executionResult = await util.executeAndThrowIfError(`az containerapp env list -g ${resourceGroup} --query [0].name`);
```
cormacpayne, line 244: nit: can we move the command that we're executing to its own `command` variable?

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    /**
+     * Gets the default location for the Container App provider.
+     * @returns the default location if found, otherwise 'eastus2'
+     */
+    public async getDefaultContainerAppLocation(): Promise<string> {
+        toolHelper.writeDebug(`Attempting to get the default location for the Container App service for the subscription.`);
+        try {
+            let command = `az provider show -n Microsoft.App --query \"resourceTypes[?resourceType=='containerApps'].locations[] | [0]\"`
```
cormacpayne, line 210: nit: do we need to escape the double quotes in this string? I'm afraid that it'll interpreted as `\"` instead of `"` since we're using the tick mark to wrap the entire string variable rather than double quotes

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,434 @@
...
+     * Determines the runtime stack to use for the runnable application image.
+     * @param appSourcePath - the path to the application source on the machine
+     * @returns a string representing the runtime stack that can be used for the Oryx MCR runtime images
+     */
+    public async determineRuntimeStackAsync(appSourcePath: string): Promise<string> {
+        toolHelper.writeDebug('Attempting to determine the runtime stack needed for the provided application source');
+        try {
+            let dockerTool: string = await toolHelper.which("docker", true);
+            // Use 'oryx dockerfile' command to determine the runtime stack to use and write it to a temp file
+            let command = `run --rm -v ${appSourcePath}:/app ${ORYX_CLI_IMAGE} /bin/bash -c \"oryx dockerfile /app | head -n 1 | sed 's/ARG RUNTIME=//' >> /app/oryx-runtime.txt\"`
```
cormacpayne, line 364: nit: same comment here about escaping the double quotes

src/ContainerAppHelper.ts
```diff
@@ -0,0 +1,434 @@
...
+            let command: string = '';
+            let commandLine = '';
+            if (IS_WINDOWS_AGENT) {
+                let packZipDownloadUri: string = 'https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-windows.zip';
+                let packZipDownloadFilePath: string = path.join(PACK_CMD, 'pack-windows.zip');
+                command = `New-Item -ItemType Directory -Path ${PACK_CMD} -Force | Out-Null; Invoke-WebRequest -Uri ${packZipDownloadUri} -OutFile ${packZipDownloadFilePath}; Expand-Archive -LiteralPath ${packZipDownloadFilePath} -DestinationPath ${PACK_CMD}; Remove-Item -Path ${packZipDownloadFilePath}`;
+                commandLine = 'pwsh';
+            } else {
+                let tgzSuffix = os.platform() == 'darwin' ? 'macos' : 'linux';
+                command = `(curl -sSL \"https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-${tgzSuffix}.tgz\" | ` +
```
cormacpayne, line 424: nit: same comment here about escaping the double quotes -- I know this one is escape in the Azure DevOps Task, but I'm pretty sure it doesn't need to be since we're using tick marks to wrap the string (I might have been using double quotes to wrap the string previously and forgot to change)

src/ContainerRegistryHelper.ts
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Authorizes Docker to make calls to the provided ACR instance using an access token that is generated via
+     * the 'az acr login --expose-token' command.
+     * @param acrName - the name of the ACR instance to authenticate calls to.
+     */
+    public async loginAcrWithAccessTokenAsync(acrName: string) {
+        toolHelper.writeDebug(`Attempting to log in to ACR instance "${acrName}" with access token`);
+        try {
+            let commandLine = os.platform() === 'win32' ? 'pwsh' : 'bash';
+            await util.executeAndThrowIfError(`${commandLine} -c "CA_ADO_TASK_ACR_ACCESS_TOKEN=$(az acr login --name ${acrName} --output json --expose-token --only-show-errors | jq -r '.accessToken'); docker login ${acrName}.azurecr.io -u 00000000-0000-0000-0000-000000000000 -p $CA_ADO_TASK_ACR_ACCESS_TOKEN > /dev/null 2>&1"`);
```
cormacpayne, line 35: Any reason we need to define a `dockerTool` variable for the other two functions when calling the `docker` command, but for this command that we're setting up, we're just calling `docker` directly?

src/GithubActionsToolHelper.ts
```diff
@@ -0,0 +1,89 @@
...
+        return containerAppName;
+    }
+
+    public getTelemetryArg(): string {
+        return `CALLER_ID=github-actions-v1`;
+    }
+
+    public getEventName(): string {
+        return `ContainerAppsGitHubActionV1`;
+    }
```
cormacpayne, line 88: I'm thinking we should change these both to end with `v2` and `V2`, respectively, since we'll be creating a `v2` tag for the release associated with these changes.

src/GithubActionsToolHelper.ts
```diff
@@ -0,0 +1,89 @@
...
+              stderr: stderr
+            }
+            resolve(executionResult);
+          });
+        }catch(err){
+            throw err;
+        }
+    }
+
+    public getInput(name: string, options?: core.InputOptions): string {
```
cormacpayne, line 62: Rather than this function accepting an `options?: core.InputOptions` argument, which is specific to the `@actions/core` library, we should change the argument to be `required?: boolean` and create our own option that uses this `required` argument when we call `core.getInput` below

cormacpayne (commented):
Last set of minor comments and then I think we're good 😃

daniv-msft:
I was SURE I made this comment on the PR a couple of weeks back, and I cannot find it so I must have dreamt/forgot to publish it: is it right to assume that most of the code changes are exactly the same as the ADO Task, and that as such we should focus the review on what has changed? If yes, could you please point out the files that are new additions?

azurecontainerapps.ts
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    /**
+     * Builds a runnable application image using a Dockerfile or the builder and pushes it to ACR.
+     */
+    private static async buildAndPushImageAsync() {
+        // Get the name of the image to build if it was provided, or generate it from build variables
+        this.imageToBuild = this.toolHelper.getInput('imageToBuild', { required: false });
+        if (this.util.isNullOrEmpty(this.imageToBuild)) {
+            this.imageToBuild = `${this.acrName}.azurecr.io/gh-action/container-app:${this.buildId}.${this.buildNumber}`;
```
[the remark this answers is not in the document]

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
snehapar9: Removed double quotes.

src/ContainerAppHelper.ts
snehapar9: Removed double quotes.

src/ContainerRegistryHelper.ts
snehapar9: We do not need `dockerTool` in a separate variable. We are able to call `docker` directly.

src/GithubActionsToolHelper.ts
snehapar9: Updated.

src/GithubActionsToolHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
snehapar9: Updated.

snehapar9:
> I was SURE I made this comment on the PR a couple of weeks back, and I cannot find it so I must have dreamt/forgot to publish it: is it right to assume that most of the code changes are exactly the same as the ADO Task, and that as such we should focus the review on what has changed? If yes, could you please point out the files that are new additions?

[From here](https://github.com/Azure/container-apps-deploy-pipelines-task/tree/main/azurecontainerapps/src) all the files under src should be the same with some expections `AzureAuthenticationHelper.ts` is removed in Github Actions because it is not required while using the actions toolkit. `GitHubActionsToolHelper.ts` has been newly added. It contains functionality unique to just the Github Actions toolkit, the plan is to create a similar file in the Azure Devops Task to house functionalities unique to the Azure Devops lib.

The dist folder is a new addition but was created by compiling azurecontainerapps.ts and does not need to be reviewed.

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
```diff
@@ -359,10 +360,9 @@ export class ContainerAppHelper {
     public async determineRuntimeStackAsync(appSourcePath: string): Promise<string> {
         toolHelper.writeDebug('Attempting to determine the runtime stack needed for the provided application source');
         try {
-            let dockerTool: string = await toolHelper.which("docker", true);
             // Use 'oryx dockerfile' command to determine the runtime stack to use and write it to a temp file
-            let command = `run --rm -v ${appSourcePath}:/app ${ORYX_CLI_IMAGE} /bin/bash -c \"oryx dockerfile /app | head -n 1 | sed 's/ARG RUNTIME=//' >> /app/oryx-runtime.txt\"`
-            await util.executeAndThrowIfError(`${dockerTool} ${command}`)
+            let command = `docker run --rm -v ${appSourcePath}:/app ${ORYX_CLI_IMAGE} /bin/bash -c oryx dockerfile /app | head -n 1 | sed 's/ARG RUNTIME=//' >> /app/oryx-runtime.txt`
```
cormacpayne: Need double quotes here

```typescript
docker run --rm -v ${appSourcePath}:/app ${ORYX_CLI_IMAGE} /bin/bash -c "oryx dockerfile /app | head -n 1 | sed 's/ARG RUNTIME=//' >> /app/oryx-runtime.txt"
```

src/ContainerAppHelper.ts
```diff
@@ -421,7 +421,7 @@ export class ContainerAppHelper {
                 commandLine = 'pwsh';
             } else {
                 let tgzSuffix = os.platform() == 'darwin' ? 'macos' : 'linux';
-                command = `(curl -sSL \"https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-${tgzSuffix}.tgz\" | ` +
+                command = `(curl -sSL https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-${tgzSuffix}.tgz | ` +
```
cormacpayne: I _think_ we need double quotes here, but we can try without

src/TelemetryHelper.ts
```diff
@@ -86,9 +86,7 @@ export class TelemetryHelper {
                 }
 
                 let eventName = toolHelper.getEventName();
-                let dockerTool: string = await toolHelper.which("docker", true);
-
-                await util.executeAndThrowIfError(`${dockerTool} run --rm ${ORYX_CLI_IMAGE} /bin/bash -c oryx telemetry --event-name ${eventName} --processing-time ${taskLengthMilliseconds} ${resultArg} ${scenarioArg} ${errorMessageArg}`);
+                await util.executeAndThrowIfError(`docker run --rm ${ORYX_CLI_IMAGE} /bin/bash -c oryx telemetry --event-name ${eventName} --processing-time ${taskLengthMilliseconds} ${resultArg} ${scenarioArg} ${errorMessageArg}`);
```
cormacpayne, line 89: Need double quotes here

```typescript
docker run --rm ${ORYX_CLI_IMAGE} /bin/bash -c "oryx telemetry --event-name ${eventName} --processing-time ${taskLengthMilliseconds} ${resultArg} ${scenarioArg} ${errorMessageArg}"
```

src/ContainerRegistryHelper.ts
```diff
@@ -32,7 +31,7 @@ export class ContainerRegistryHelper {
         toolHelper.writeDebug(`Attempting to log in to ACR instance "${acrName}" with access token`);
         try {
             let commandLine = os.platform() === 'win32' ? 'pwsh' : 'bash';
-            await util.executeAndThrowIfError(`${commandLine} -c "CA_ADO_TASK_ACR_ACCESS_TOKEN=$(az acr login --name ${acrName} --output json --expose-token --only-show-errors | jq -r '.accessToken'); docker login ${acrName}.azurecr.io -u 00000000-0000-0000-0000-000000000000 -p $CA_ADO_TASK_ACR_ACCESS_TOKEN > /dev/null 2>&1"`);
+            await util.executeAndThrowIfError(`${commandLine} -c CA_ADO_TASK_ACR_ACCESS_TOKEN=$(az acr login --name ${acrName} --output json --expose-token --only-show-errors | jq -r '.accessToken'); docker login ${acrName}.azurecr.io -u 00000000-0000-0000-0000-000000000000 -p $CA_ADO_TASK_ACR_ACCESS_TOKEN > /dev/null 2>&1`);
```
cormacpayne, line 34: Need double quotes here:

```typescript
${commandLine} -c "CA_ADO_TASK_ACR_ACCESS_TOKEN=$(az acr login --name ${acrName} --output json --expose-token --only-show-errors | jq -r '.accessToken'); docker login ${acrName}.azurecr.io -u 00000000-0000-0000-0000-000000000000 -p $CA_ADO_TASK_ACR_ACCESS_TOKEN > /dev/null 2>&1"
```

src/ContainerAppHelper.ts
```diff
@@ -241,7 +241,8 @@ export class ContainerAppHelper {
     public async getExistingContainerAppEnvironment(resourceGroup: string) {
         toolHelper.writeDebug(`Attempting to get the existing Container App Environment in resource group "${resourceGroup}"`);
         try {
-            let executionResult = await util.executeAndThrowIfError(`az containerapp env list -g ${resourceGroup} --query [0].name`);
+            let command = `az containerapp env list -g ${resourceGroup} --query [0].name`
```
cormacpayne, line 244: Let's add double quotes around the `--query` argument here just to be safe

```typescript
az containerapp env list -g ${resourceGroup} --query "[0].name"
```

cormacpayne (approved):
LGTM :shipit:

daniv-msft:
> > I was SURE I made this comment on the PR a couple of weeks back, and I cannot find it so I must have dreamt/forgot to publish it: is it right to assume that most of the code changes are exactly the same as the ADO Task, and that as such we should focus the review on what has changed? If yes, could you please point out the files that are new additions?
> 
> [From here](https://github.com/Azure/container-apps-deploy-pipelines-task/tree/main/azurecontainerapps/src) all the files under src should be the same with some expections `AzureAuthenticationHelper.ts` is removed in Github Actions because it is not required while using the actions toolkit. `GitHubActionsToolHelper.ts` has been newly added to this repo. It contains functionality unique to just the Github Actions toolkit, the plan is to create a similar file in the Azure Devops Task to house functionalities unique to the Azure Devops lib.
> 
> The dist folder is a new addition but was created by compiling azurecontainerapps.ts and does not need to be reviewed.

Thanks @snehapar9! I'm reviewing specifically `GitHubActionsToolHelper.ts` in that case.

daniv-msft (approved):

daniv-msft (approved):

daniv-msft (approved):

azurecontainerapps.ts
```diff
@@ -26,6 +26,11 @@ export class azurecontainerapps {
                 await this.authenticateAzureContainerRegistryAsync();
             }
 
+            // If a Container Registry URL was provided, try to authenticate against it
+            if (!this.util.isNullOrEmpty(this.registryUrl)) {
+                await this.authenticateContainerRegistryAsync();
+            }
```
cormacpayne, line 32: Let's move this block before the `acrName` check block above because `authenticateAzureContainerRegistryAsync` is going to set the value of `registryUrl` to include the value of `acrName`, so both of these if-statements will be hit

cormacpayne (approved):
:shipit:

PR #63, opened 2023-10-02
Update builder image and support multiple stacks

cormacpayne:
This pull request aims to introduce the following changes:
- Update the Oryx++ builder image used to the `20231004.1` tag
- Support multiple stacks of the Oryx++ builder (Debian bookworm and bullseye) and try to build an application with each stack until one is successful
- Introduce the `builderStack` input to allow the user to specify a specific tag of the builder image they'd like to use or a specific OS stack


README.md
```diff
@@ -161,15 +161,16 @@ For more information on the structure of the YAML configuration file, please vis
...
 | `runtimeStack`            | No       | The platform version stack used in the final runnable application image that is deployed to the Container App. The value should be provided in the formation `<platform>:<version>`. If not provided, this value is determined by Oryx based on the contents of the provided application. Please refer to [this document](https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md) for more information on supported runtime stacks for Oryx. |
+| `builderVersion`          | No       | The version of the Oryx++ builder that should be used to build the provided application source and produce the runnable application image. You can provide a specific image tag, such as "debian-bookworm-20230928.2", or you can provide a supported OS version, such as "debian-bookworm" or "debian-bullseye", and the current supported image tag for the corresponding OS version will be used. If no version is provided, this action will attempt to build the provided application source with each supported OS version until a successful build is completed. |
```
daniv-msft, line 173: I'm fine if we keep the current `builderVersion` name. I'm trying to determine if something like `builderStack` might be better? I'm not feeling strongly and I recognize it might create some confusion as we're not advertising that concept much outside.

cormacpayne: Yeah I actually thought about renaming the input to `builderStack` as I was typing out the description of the PR 😆 I think we should change it to be more clear about what it's representing, which is a stack.

daniv-msft: OK sounds good!

action.yml
```diff
@@ -439,7 +426,8 @@ runs:
       if: ${{ env.CA_GH_ACTION_INGRESS_ENABLED == 'true' && inputs.targetPort == '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' && inputs.appSourcePath != '' && env.CA_GH_ACTION_YAML_PROVIDED != 'true' }}
       shell: bash
       run: |
-        if [[ "${{ env.CA_GH_ACTION_RUNTIME_STACK }}" == "python:"* ]]; then echo "CA_GH_ACTION_TARGET_PORT=80" >> $GITHUB_ENV; else echo "CA_GH_ACTION_TARGET_PORT=8080" >> $GITHUB_ENV; fi
+        CA_GH_ACTION_TARGET_PORT="8080"
+        echo "CA_GH_ACTION_TARGET_PORT=${CA_GH_ACTION_TARGET_PORT}" >> $GITHUB_ENV
```
daniv-msft, line 430: Could you please clarify where the Python exception related to port 80 moved? Did we not need it anymore?

action.yml
```diff
@@ -502,15 +495,68 @@ runs:
...
+          echo "CA_GH_ACTION_USE_BUSTER_BUILDER=true" >> $GITHUB_ENV
+
+    - name: Create runnable application image using buster Oryx++ Builder
+      if: ${{ inputs.appSourcePath != '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' && ((inputs.builderVersion == '' && env.CA_GH_ACTION_USE_BUSTER_BUILDER == 'true') || inputs.builderVersion == 'debian-buster') }}
+      shell: bash
+      run: |
+        pack build \
+          ${{ env.CA_GH_ACTION_IMAGE_TO_DEPLOY }} \
+          --path ${{ inputs.appSourcePath }} \
+          --builder mcr.microsoft.com/oryx/builder:debian-buster-20230928.2 \
```
daniv-msft, line 547: Do we even need Buster? I don't think we have any new language version that would come with it? If not, I'd rather remove it so that we don't need to remove it at some point in the future when we're not sure if we can remove it. :)

action.yml
```diff
@@ -502,15 +495,68 @@ runs:
...
+        echo "CA_GH_ACTION_RUNTIME_STACK_VERSION=${CA_GH_ACTION_RUNTIME_STACK_VERSION}" >> $GITHUB_ENV
+
+    - name: Set environment variables for builder invocation
+      if: ${{ inputs.appSourcePath != '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' && inputs.runtimeStack != '' }}
+      shell: bash
+      run: |
+        CA_GH_ACTION_ORYX_BUILDER_ENV_ARG="--env ORYX_PLATFORM_NAME=${{ env.CA_GH_ACTION_RUNTIME_STACK_PLATFORM }} --env ORYX_PLATFORM_VERSION=${{ env.CA_GH_ACTION_RUNTIME_STACK_VERSION }}"
+        echo "CA_GH_ACTION_ORYX_BUILDER_ENV_ARG=${CA_GH_ACTION_ORYX_BUILDER_ENV_ARG}" >> $GITHUB_ENV
+
+    - name: Create runnable application image using bookworm Oryx++ Builder
```
daniv-msft, line 516: Could you please clarify what is the upfront cost to have the bookworm builder first for all other platforms? Would you see any issues with keeping the bullseye one first?

cormacpayne: The difference between buster and bullseye is a few .NET versions that aren't supported anymore, so I'm OK with ditching buster from the selection we have here.

action.yml
cormacpayne: The difference between buster and bullseye is a few .NET versions that aren't supported anymore, so I'm OK with ditching buster from the selection we have here.

action.yml
cormacpayne: OK I knew there was something I forgot to clean-up 😄 The Oryx runtime images we use have a default port that they listen on if one wasn't provided as an environment variable, so if the user didn't provide a port for their Container App, we defaulted to the port that was defined on the platform's corresponding runtime image. Since we can update the port in the builder via the `ORYX_RUNTIME_PORT` environment variable now, we can just set a universal default of 8080 and provide that to this environment variable and not have it matter what platform was used.

action.yml
daniv-msft: Thanks. Yes giving the best perf to the highest number of customers seems like the best compromise from my perspective.

action.yml
daniv-msft: OK thanks. Yes let's remove buster and we can reevaluate if customers ask for these versions.

action.yml
daniv-msft: OK makes sense. Thanks for clarifying!

action.yml
```diff
@@ -502,15 +495,68 @@ runs:
...
+      if: ${{ inputs.appSourcePath != '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' && ((inputs.builderVersion == '' && env.CA_GH_ACTION_USE_BULLSEYE_BUILDER == 'true') || inputs.builderVersion == 'debian-bullseye') }}
+      shell: bash
+      run: |
+        pack build \
+          ${{ env.CA_GH_ACTION_IMAGE_TO_DEPLOY }} \
+          --path ${{ inputs.appSourcePath }} \
+          --builder mcr.microsoft.com/oryx/builder:debian-bullseye-20230928.2 \
+          ${{ env.CA_GH_ACTION_ORYX_BUILDER_ENV_ARG }} \
+          ${{ env.CA_GH_ACTION_ORYX_BUILDER_TELEMETRY_ARG }} || \
+          echo "CA_GH_ACTION_USE_BUSTER_BUILDER=true" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
cormacpayne: Not in this case -- we're calling `pack build` with the bullseye builder, and if it fails, the `|| echo "CA_GH_ACTION_USE_BUSTER_BUILDER=true >> $GITHUB_ENV` command is invoked, which is just setting that variable to `true`, which is then used in the next step to check whether or not we should use the buster builder. This allows us to safely say "either the bullseye builder succeeded and we don't need to try any other builders or the bullseye builder failed and we need to try the next one in line".

As a note: usage of the buster builder will be removed in the latest commit per Dan's review comment earlier.

action.yml
```diff
@@ -502,15 +495,68 @@ runs:
...
+      if: ${{ inputs.appSourcePath != '' && env.CA_GH_ACTION_DOCKERFILE_PATH == '' && (inputs.builderVersion == '' || inputs.builderVersion == 'debian-bookworm') }}
+      shell: bash
+      run: |
+        pack build \
+          ${{ env.CA_GH_ACTION_IMAGE_TO_DEPLOY }} \
+          --path ${{ inputs.appSourcePath }} \
+          --builder mcr.microsoft.com/oryx/builder:debian-bookworm-20230928.2 \
+          ${{ env.CA_GH_ACTION_ORYX_BUILDER_ENV_ARG }} \
+          ${{ env.CA_GH_ACTION_ORYX_BUILDER_TELEMETRY_ARG }} || \
+          echo "CA_GH_ACTION_USE_BULLSEYE_BUILDER=true" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
cormacpayne: Same comment above applies here :shipit:

action.yml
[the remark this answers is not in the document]
cormacpayne: I changed this logic a little so that the default port used across the board is `80`, but I don't think it necessarily matters what we're using now since we'll be telling the Oryx runtime image that we're targeting this port (or the port provided by the user), and it'll update its internal environment variables to account for that.

daniv-msft (approved):

william-msft (approved):

daniv-msft (approved):

daniv-msft (dismissed):

daniv-msft (approved):

PR #69, opened 2023-11-01
Support private registry scenario

snehapar9:
Add support for private registry - https://msazure.visualstudio.com/Antares/_sprints/taskboard/ContainerApps/Antares/Germanium/CY23Q4?workitem=24827107


azurecontainerapps.ts
```diff
@@ -150,8 +154,12 @@ export class azurecontainerapps {
         // Get the YAML configuration file, if provided
         this.yamlConfigPath = this.toolHelper.getInput('yamlConfigPath', false) as string;
 
+        // Get the name of the image to build, if provided
+        this.imageToBuild = this.toolHelper.getInput('imageToBuild', false) as string;
```
cormacpayne, line 158: Aren't we already reading the value for this input elsewhere in the code?

azurecontainerapps.ts
```diff
@@ -150,8 +154,12 @@ export class azurecontainerapps {
...
+
         // Ensure that acrName or registryUrl is also provided if appSourcePath is provided
-        if (!this.util.isNullOrEmpty(this.appSourcePath) && this.util.isNullOrEmpty(this.acrName) && this.util.isNullOrEmpty(this.registryUrl)) {
+        // and if imageToBuild is either empty or doesn't start with 'default/' (i.e. it's not source to cloud scenario)
+        if (!this.util.isNullOrEmpty(this.appSourcePath) && this.util.isNullOrEmpty(this.acrName) && this.util.isNullOrEmpty(this.registryUrl) && (this.util.isNullOrEmpty(this.imageToBuild) || !this.imageToBuild.startsWith('default/'))) {
```
cormacpayne, line 162: Do we need to check that `imageToBuild` is empty? Isn't it sufficient to just check that the value doesn't begin with `default/`? Maybe it depends on the value returned by `toolHelper.getInput` if no value for the input is provided, the string may be `""` or `null`, I'm not sure off the top of my head.

cormacpayne, line 162: How would users not going through the Portal UX leverage the source-to-cloud functionality in the Azure CLI? It seems like they would need to know to provide a random `default/*` value for `imageToBuild` or else this validation will fail their workflow.

azurecontainerapps.ts
```diff
@@ -150,8 +154,12 @@ export class azurecontainerapps {
...
             let missingRegistryUrlMessage = `The 'acrName' or 'registryUrl' argument must be provided when the 'appSourcePath' argument is provided.`;
```
cormacpayne, line 163: We should update this error message to reflect the change in our validation here.

azurecontainerapps.ts
```diff
@@ -170,6 +178,20 @@ export class azurecontainerapps {
             this.toolHelper.writeError(conflictingArgumentsMessage);
             throw Error(conflictingArgumentsMessage);
         }
+
+        // Ensure that the imageToBuild and source arguments are both provided and imageToBuild starts with 'default/'
+        if (!this.util.isNullOrEmpty(this.imageToBuild) && !this.util.isNullOrEmpty(this.appSourcePath) && !this.imageToBuild.startsWith('default/')) {
```
cormacpayne, line 183: I'm not sure I understand this check -- if the user wants to use any non-internal registry flow, this validation automatically breaks them. For example, if they provide their application source with a Dockerfile, and they provide a value for `imageToBuild` that tells us "I want to build the Dockerfile in my provided application source with this image name", we're going to fail them here. Same goes for basically every non-internal registry flow where the user tells us what they want their image name to be.

azurecontainerapps.ts
snehapar9: Users should be able to trigger source to cloud from the CLI by executing `az containerapp up --source ...` by meeting these conditions. 
1. No Docker running/installed.
2. No registry available/provided.

They don't need to provide a random `default/*` image. This image is needed only for triggering cloud build from source control.

azurecontainerapps.ts
snehapar9: Ok, this condition was meant to be `if (!this.util.isNullOrEmpty(this.imageToBuild) && !this.util.isNullOrEmpty(this.appSourcePath) && !this.imageToBuild.startsWith('default/')`. 
Didn't realize there was a ! after `this.imageToBuild.startsWith('default/')`.

This condition makes sure that the `appSourcePath` arg is provided for the internal registry scenario.

azurecontainerapps.ts
cormacpayne: Let me rephrase the question: if the user wants to use the cloud build service with an internal registry, they are providing their application source and _are not_ providing an ACR or external registry, so their workflow will fail due to this validation in all cases _unless_ they provide a value for `imageToBuild` that is `default/*`, right? That's how this conditional seems to operate based on my interpretation.

azurecontainerapps.ts
cormacpayne: You'll also want to remove the `!` before the check for `appSourcePath` because right now the conditional is equivalent to "if the user provides an application source path AND they provide an image to build that begins with `default/`, throw an exception"

azurecontainerapps.ts
snehapar9: Thank you for clarifying. Sorry I misunderstood your question.

I hope I'm not missing anything else, but the source control API should supply the `imageToBuild` arg starting with `default/* in the template that it sets up while calling the action`?

That's how this action will be called? So here is how the flow looks like from my understanding - 

Portal
Customer creates container app/sets source control by choosing `Push to internal registry` -> Source Control API performs validation (skips mandatory check for registry because image matches `default/*` -> Calls action from the template and supplies the `imageToBuild` arg starting with `default/*` so this isn't something that is set by the customer.

azurecontainerapps.ts
snehapar9: I missed that. That you!

azurecontainerapps.ts
cormacpayne: > I hope I'm not missing anything else, but the source control API should supply the `imageToBuild` arg starting with `default/*` in the template that it sets up while calling the action?

This is correct, but 

> If customers were to use the internal registry, they would directly just run the CLI commands?

I don't think this is correct -- there are a lot of customers who use the action independent of the source control API, so we need to make sure that if any of those customers want to use the internal registry functionality that they can do so without needing to go through the source control API to provide them a `default/*` value or without telling them that they need to use the CLI themselves. Does that makes sense?

azurecontainerapps.ts
```diff
@@ -31,8 +31,11 @@ export class azurecontainerapps {
                 await this.authenticateAzureContainerRegistryAsync();
             }
 
+            // Determine if the image should be built and pushed using the CLI
+            this.useCLIToBuildAndPushImage = !this.util.isNullOrEmpty(this.appSourcePath) && ((!this.util.isNullOrEmpty(this.registryUrl) && this.registryUrl.endsWith('.azurecr.io')) || (this.util.isNullOrEmpty(this.registryUrl) && this.imageToBuild.startsWith('default/')))
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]

azurecontainerapps.ts
snehapar9: We don't need this anymore because we removed the validations for `imageToBuild` arg.

azurecontainerapps.ts
snehapar9: Removed this vaildation.

snehapar9: Synced offline. We don't need to validate against `imageToBuild` in the GitHub action because if `appSourcePath` is provided and both `acrName` and `registryUrl` are Null, we default to the internal registry flow.

azurecontainerapps.ts
snehapar9: Removed this validation.

azurecontainerapps.ts
```diff
@@ -150,12 +157,8 @@ export class azurecontainerapps {
...
         this.yamlConfigPath = this.toolHelper.getInput('yamlConfigPath', false) as string;
 
-        // Ensure that acrName or registryUrl is also provided if appSourcePath is provided
-        if (!this.util.isNullOrEmpty(this.appSourcePath) && this.util.isNullOrEmpty(this.acrName) && this.util.isNullOrEmpty(this.registryUrl)) {
-            let missingRegistryUrlMessage = `The 'acrName' or 'registryUrl' argument must be provided when the 'appSourcePath' argument is provided.`;
-            this.toolHelper.writeError(missingRegistryUrlMessage);
-            throw Error(missingRegistryUrlMessage);
-        }
+        // Get the name of the image to build if it was provided, or generate it from build variables
+        this.imageToBuild = this.toolHelper.getInput('imageToBuild', false);
```
daniv-msft, line 161: Could you please clarify why we needed to move this one out of `buildAndPushImageAsync`?

src/Utility.ts
```diff
@@ -20,6 +20,7 @@ export class Utility {
    */
   public async setAzureCliDynamicInstall() {
     await this.execute(`az config set extension.use_dynamic_install=yes_without_prompt`);
```
cormacpayne, line 22: nit: we shouldn't need this command anymore if we're explicitly installing the `containerapp` extension below

src/Utility.ts
```diff
@@ -20,6 +20,7 @@ export class Utility {
...
+    await this.execute(`az extension add --name containerapp --version 0.3.43`);
```
cormacpayne, line 23: Do we need to snap to a specific version of the extension? I think it would be best if we always used the latest so we didn't need to release a new version of the Action every time there's a new extension release with a fix that we need.

src/ContainerAppHelper.ts
```diff
@@ -30,11 +30,10 @@ export class ContainerAppHelper {
         containerAppName: string,
         resourceGroup: string,
         environment: string,
-        imageToDeploy: string,
         optionalCmdArgs: string[]) {
         toolHelper.writeDebug(`Attempting to create Container App with name "${containerAppName}" in resource group "${resourceGroup}" based from image "${imageToDeploy}"`);
```
cormacpayne, line 33: `imageToDeploy` is still referenced in this debug message even though we removed the argument

src/ContainerAppHelper.ts
```diff
@@ -75,11 +74,10 @@ export class ContainerAppHelper {
     public async updateContainerApp(
         containerAppName: string,
         resourceGroup: string,
-        imageToDeploy: string,
         optionalCmdArgs: string[]) {
         toolHelper.writeDebug(`Attempting to update Container App with name "${containerAppName}" in resource group "${resourceGroup}" based from image "${imageToDeploy}"`);
```
cormacpayne, line 76: Same comment here about `imageToDeploy` still being referenced

src/ContainerAppHelper.ts
```diff
@@ -94,21 +90,20 @@ export class ContainerAppHelper {
      * Updates an existing Azure Container App using the 'az containerapp up' command.
      * @param containerAppName - the name of the existing Container App
      * @param resourceGroup - the resource group that the existing Container App is found in
-     * @param imageToDeploy - the name of the runnable application image that the Container App will be based from
      * @param optionalCmdArgs - a set of optional command line arguments
      * @param ingress - the ingress that the Container App will be exposed on
      * @param targetPort - the target port that the Container App will be exposed on
+     * @param appSourcePath - the path to the application source on the machine
```
cormacpayne, line 96: Unintentionally added?

azurecontainerapps.ts
```diff
@@ -511,6 +523,14 @@ export class azurecontainerapps {
                 this.commandLineArgs.push(`--env-vars ${environmentVariables}`);
             }
         }
+
+        if(!this.imageToDeploy.startsWith('default/')) {
```
cormacpayne, line 527: nit: let's make sure there's a space between the `if` and the opening `(` of the conditional

(same comment for the if-statement added a couple of lines below)

azurecontainerapps.ts
```diff
@@ -343,16 +349,17 @@ export class azurecontainerapps {
...
+        * file is provided.
+        */
+    private static setupContainerAppImageProperties() {
         // Get the name of the image to build if it was provided, or generate it from build variables
         this.imageToBuild = this.toolHelper.getInput('imageToBuild', false);
+
         if (this.util.isNullOrEmpty(this.imageToBuild)) {
             const imageRepository = this.toolHelper.getDefaultImageRepository()
-            this.imageToBuild = `${this.registryUrl}/${imageRepository}:${this.buildId}.${this.buildNumber}`;
+            this.imageToBuild = this.util.isNullOrEmpty(this.registryUrl) ? `default/${imageRepository}:${this.buildId}.${this.buildNumber}` : `${this.registryUrl}/${imageRepository}:${this.buildId}.${this.buildNumber}`;
```
cormacpayne, line 362: Do we use this `imageToBuild` value anywhere for the internal registry scenario? I haven't checked everywhere yet, but it seems like the main thing we use it for is to set the value of `useInternalRegistry`, but it seems like for that case we really are just checking if `this.registryUrl` is null or empty (since we set this `imageToBuild` value to `default/*` if no registry was provided by the user)

azurecontainerapps.ts
snehapar9: We are leveraging the `imageToBuild` arg to determine if we want to use the CLI to build the image or this `buildAndPushImageAsync` method.  We need to move this out of this method so we can decide if we want to execute this method based on the condition below.

image

src/Utility.ts
snehapar9: Added this as a temporary fix until `--source` GA's. But I'll change this so we always install the latest version.

src/ContainerAppHelper.ts
snehapar9: Removed `imageToDeploy` from the debug message.

src/ContainerAppHelper.ts
snehapar9: Removed `imageToDeploy` from here.

src/ContainerAppHelper.ts
snehapar9: yes, thanks

azurecontainerapps.ts
snehapar9: That is correct! But for update scenarios, I understand the CLI may have context of the existing container app's registry info but the action would not?
so, we need this `imageToBuild` arg to determine if we want to add `--source` to the `az containerapp update` command?

cormacpayne (approved):

src/Utility.ts
```diff
@@ -15,11 +15,10 @@ export class Utility {
   }
 
   /**
-   * Sets the Azure CLI to dynamically install extensions that are missing. In this case, we care about the
-   * Azure Container Apps module being dynamically installed while it's still in preview.
+   * Sets the Azure CLI to dynamically install extensions that are missing.
    */
   public async setAzureCliDynamicInstall() {
```
daniv-msft, line 20: Nit: not for this PR, but this method name isn't super meaningful anymore

README.md
```diff
@@ -171,7 +171,6 @@ For more information on the structure of the YAML configuration file, please vis
...
 | `builderStack`            | No       | The stack (OS) that should be used to build the provided application source and produce the runnable application image. You can provide a specific image tag for the stack, such as "debian-bookworm-20231004.1", or you can provide a supported stack name, such as "debian-bookworm" or "debian-bullseye", and the latest supported image tag for that stack will be used. If no stack is provided, this action will attempt to build the provided application source with each supported stack until there's a successful build. |
-| `buildArguments`          | No       | A list of build arguments provided as KEY=VALUE pairings and are comma-separated. If a Dockerfile has been provided or is discovered in the application source, each build argument will be passed to the `docker build` command via the `--build-arg` flag. If the Oryx++ builder is used to create a runnable application image, each build argument will be passed to the `pack build` command via the `--env` flag. |
```
daniv-msft, line 174: Should `imageToBuild` be removed as well? Or do we still use it in some cases?

.github/workflows/run-validation.yml
```diff
@@ -59,10 +59,45 @@ jobs:
         shell: bash
         run: az containerapp delete -n ${{ env.TEST_CONTAINER_APP_NAME }} -g ${{ vars.TEST_RESOURCE_GROUP_NAME }} -y
 
-      - name: Delete pushed image
```
daniv-msft, line 62: Isn't this pushed image deletion still required for the previous test? Wouldn't we leave an image in the container registry there?

daniv-msft: OK I see that we removed that logic in multiple places. Could you please share more details on why we don't need this step anymore?

README.md
snehapar9: We are still using `imageToBuild` in this PR, we may be able to remove it later but that would involve some code changes.

.github/workflows/run-validation.yml
snehapar9: Since the image is created by the CLI, we would not be able to delete the image from our gha

src/Utility.ts
snehapar9: > Thanks Dan! I've made a note of this.

.github/workflows/run-validation.yml
```diff
@@ -311,43 +335,28 @@ jobs:
...
 
-      - name: Delete pushed image
-        if: ${{ always() }}
-        shell: bash
-        run: az acr repository delete -n ${{ vars.TEST_ACR_NAME }} -t ${{ env.TEST_IMAGE_REPOSITORY }}:${{ env.TEST_IMAGE_TAG }} -y
-
-  create-using-image-yaml-linux:
-    name: 'Create app using image with YAML configuration on Linux runner'
-    runs-on: ubuntu-latest
+  create-using-image-yaml-windows:
```
daniv-msft, line 352: Could you please confirm where do these changes come from? Do we need them in this PR?

daniv-msft: Is this just coming from the revert of @cormacpayne's PR?

snehapar9: Yes, I think so

azurecontainerapps.ts
```diff
@@ -342,16 +350,18 @@ export class azurecontainerapps {
...
+    private static setupContainerAppImageProperties() {
         // Get the name of the image to build if it was provided, or generate it from build variables
         this.imageToBuild = this.toolHelper.getInput('imageToBuild', false);
+
         if (this.util.isNullOrEmpty(this.imageToBuild)) {
             const imageRepository = this.toolHelper.getDefaultImageRepository()
-            this.imageToBuild = `${this.registryUrl}/${imageRepository}:${this.buildId}.${this.buildNumber}`;
+            // Constructs the image to build based on the provided registry URL, image repository,  build ID, and build number.
+            // If the registry URL is not provided or is empty, the default registry server is used; otherwise, the provided registry URL is used.
+            this.imageToBuild = this.util.isNullOrEmpty(this.registryUrl) ? `${this.defaultRegistryServer}${imageRepository}:${this.buildId}.${this.buildNumber}` : `${this.registryUrl}/${imageRepository}:${this.buildId}.${this.buildNumber}`;
```
daniv-msft, line 364: For the Cloud Build part, I really don't think that we'll respect the value provided there. Do you know if there are any issues if the value generated there and the one we actually get in practice are different?

.github/workflows/run-validation.yml
cormacpayne: The validation changes aren't from my PR -- @snehapar9 I thought these changes were made to update our validation to reflect the changes made in your PR

azurecontainerapps.ts
```diff
@@ -537,19 +537,30 @@ export class azurecontainerapps {
                 this.commandLineArgs.push(`--env-vars ${environmentVariables}`);
             }
         }
+
+        if (!this.imageToDeploy.startsWith(this.defaultRegistryServer)) {
+            this.commandLineArgs.push(`-i ${this.imageToDeploy}`);
```
daniv-msft, line 542: Nit: should this be in an `else` block of the other condition below? Or at least an `else if`, to make sure that both arguments can never be set together?

azurecontainerapps.ts
snehapar9: Just so I'm on the same page, is your question if we will use `imageToBuild` and `imageToDeploy` in the cloud build scenario? If that's the question -> The answer is that we will not. 
We are using these args to determine if want to delegate build + push operations to the CLI. 
If we make the change to only rely on the registry server being null, we do not have to set these default values here.

.github/workflows/run-validation.yml
daniv-msft: Thanks, so do we know what are the consequences? Is it just going to add new images forever in the ACR used for the tests? If yes we should consider fixing it after this PR, maybe by deleting the whole ACR instance and recreating from scratch every time we start new tests?

azurecontainerapps.ts
daniv-msft: Thanks, my comment was more that setting a specific value for `this.imageToBuild` just for Cloud Build feels weird if we already know that we won't respect it in the Cloud Build.
That said if it's needed for another part of the changes, I'm good with keeping it as it is.

azurecontainerapps.ts
snehapar9: That sounds possible? but need to verify it does not break existing scenarios. It may probably be a better solution set the value for `imageToDeplow` only for scenarios handled by the GitHub action? 
I can get another PR out for this change. @daniv-msft Is this something we want to be checked by today or can get another for this tomorrow?

daniv-msft (approved):

PR #72, opened 2023-11-07
Remove env as a required arg

snehapar9:
Remove `--environment` as a required argument for `up` because it is not set by the source control API template

Validation - https://github.com/Azure/container-apps-deploy-action/actions/runs/6789309970/job/18456202571

### DIFF
--- a/src/ContainerAppHelper.ts
+++ b/src/ContainerAppHelper.ts
@@ -55,19 +55,13 @@ export class ContainerAppHelper {
      public async createOrUpdateContainerAppWithUp(
         containerAppName: string,
         resourceGroup: string,
-        optionalCmdArgs: string[],
-        environment?: string) {
+        optionalCmdArgs: string[]) {
         toolHelper.writeDebug(`Attempting to create Container App with name "${containerAppName}" in resource group "${resourceGroup}"`);
         try {
             let command = `az containerapp up -n ${containerAppName} -g ${resourceGroup} -l northcentralusstage`;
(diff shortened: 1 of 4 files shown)


src/ContainerAppHelper.ts
```diff
@@ -49,21 +49,25 @@ export class ContainerAppHelper {
...
+        environment?: string) {
         toolHelper.writeDebug(`Attempting to create Container App with name "${containerAppName}" in resource group "${resourceGroup}"`);
         try {
-            let command = `az containerapp up -n ${containerAppName} -g ${resourceGroup} --environment ${environment} -l northcentralusstage`;
+            let command = `az containerapp up -n ${containerAppName} -g ${resourceGroup} -l northcentralusstage`;
             optionalCmdArgs.forEach(function (val: string) {
                 command += ` ${val}`;
             });
+
+            if (!util.isNullOrEmpty(environment)) {
```
daniv-msft, line 67: Just want to be sure I get it: is there any case where we pass this environment in the current code? If not, then I would rather remove the logic until we know for sure we need it.

daniv-msft (approved):

src/ContainerAppHelper.ts
snehapar9: We are not passing this `--environment` value to this method in our code. So, I removed it.

daniv-msft: Thanks

daniv-msft (approved):

daniv-msft (approved):

PR #75, opened 2023-11-09
Build and push non cloud scenarios from GHA

snehapar9:
This PR contains the following changes.

1. Build and push image using GHA for non-cloud build scenarios.
2. Solely rely on registry url being null to determine internal registry scenario.
3. Rename `setAzureCliDynamicInstall` to `installAzureCliExtension`
4. Ensure `--image` is not used in non-cloud build scenarios and `--source` is used only in cloud build scenarios.

### DIFF
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -529,7 +529,7 @@ export class azurecontainerapps {
             }
         }
 
-        // Ensure 'imageToDeploy' argument and '--source' argument are not both provided
+        // Ensure '-i' argument and '--source' argument are not both provided
         if (!this.util.isNullOrEmpty(this.imageToDeploy)) {
             this.commandLineArgs.push(`-i ${this.imageToDeploy}`);
         } else if (this.createOrUpdateContainerApp) {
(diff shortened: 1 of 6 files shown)


daniv-msft (approved):

daniv-msft (approved):

daniv-msft (approved):

azurecontainerapps.ts
```diff
@@ -573,15 +563,15 @@ export class azurecontainerapps {
             return;
         }
 
-        if (this.shouldUseUpdateCommand && !createOrUpdateContainerAppWithUp) {
+        if (this.shouldUseUpdateCommand && !this.shouldCreateOrUpdateContainerAppWithUp) {
```
cormacpayne, line 566: nit: we should see if the conditionals for each of these variables are specific enough to distinguish them a part; _i.e._, it would be great if this given conditional was just

```typescript
if (this.shouldUseUpdateCommand)
```

since it would imply that some conditional set for `this.shouldCreateOrUpdateContainerAppWithUp` was not met and it's `false`, so we wouldn't need the additional `&& !this.shouldCreateOrUpdateContainerAppWithUp` piece.

cormacpayne (approved):

azurecontainerapps.ts
snehapar9: Thanks @cormacpayne! Is this something we can look into post-Ignite?

PR #76, opened 2023-11-10
Remove location

snehapar9:
This PR removes hardcoded location while running `up` because cloud build is not specific to any region anymore.

### DIFF
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -232,26 +232,25 @@ export class azurecontainerapps {
     private static async getLocation(): Promise<string> {
         // Set deployment location, if provided
         let location: string = this.toolHelper.getInput('location', false);
-
         if (!this.util.isNullOrEmpty(location)) {
             return location;
         }
 
         // If no location was provided, attempt to discover the location of the existing Container App Environment linked to the Container App
         // or Container App Environment provided in the resource group or use the default location.
-        // Get the resource group if it was provided, or generate it from the Container App name
+        // Get the resource group if it was provided
         let resourceGroup: string = this.toolHelper.getInput('resourceGroup', false);
 
         if (!this.util.isNullOrEmpty(resourceGroup)) {
             // Check if Container App exists in the resource group provided and get the location from the Container App Environment linked to it
-            let doesContainerAppExist = await this.appHelper.doesContainerAppExist(this.containerAppName, resourceGroup);
-            if (doesContainerAppExist) {
+            let containerAppExists = await this.appHelper.doesContainerAppExist(this.containerAppName, resourceGroup);
+            if (containerAppExists) {
                 // Get the name of the Container App Environment linked to the Container App
                 var environmentName = await this.appHelper.getExistingContainerAppEnvironmentName(this.containerAppName, resourceGroup);
 
                 // Check if environment exists in the resource group provided and get the location
-                var doesContainerAppEnvironmentExistInResourceGroup = !this.util.isNullOrEmpty(environmentName) ? await this.appHelper.doesContainerAppEnvironmentExist(environmentName, resourceGroup) : false;
-                if (doesContainerAppEnvironmentExistInResourceGroup) {
+                var containerAppEnvironmentExistsInResourceGroup = !this.util.isNullOrEmpty(environmentName) ? await this.appHelper.doesContainerAppEnvironmentExist(environmentName, resourceGroup) : false;
+                if (containerAppEnvironmentExistsInResourceGroup) {
                     // Get the location of the Container App Environment linked to the Container App
                     location = await this.appHelper.getExistingContainerAppEnvironmentLocation(environmentName, resourceGroup);
                     return location;
@@ -261,8 +260,8 @@ export class azurecontainerapps {
             let containerAppEnvironment: string = this.toolHelper.getInput('containerAppEnvironment', false);
 
             // Check if Container App Environment is provided and exits in the resource group provided and get the location
-            let doesContainerAppEnvironmentExist = !this.util.isNullOrEmpty(containerAppEnvironment) ? await this.appHelper.doesContainerAppEnvironmentExist(containerAppEnvironment, resourceGroup) : false;
-            if (doesContainerAppEnvironmentExist) {
+            let containerAppEnvironmentExists = !this.util.isNullOrEmpty(containerAppEnvironment) ? await this.appHelper.doesContainerAppEnvironmentExist(containerAppEnvironment, resourceGroup) : false;
+            if (containerAppEnvironmentExists) {
                 location = await this.appHelper.getExistingContainerAppEnvironmentLocation(containerAppEnvironment, resourceGroup);
                 return location;
             }
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -265,7 +265,7 @@ export class azurecontainerapps {
             }
         }
 
-        // Get the default location for if the Container App or Container App Environment was not found in the resource group provided.
+        // Get the default location if the Container App or Container App Environment was not found in the resource group provided.
         location = await this.appHelper.getDefaultContainerAppLocation();
         return location;
 
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -239,23 +239,30 @@ export class azurecontainerapps {
          // or Container App Environment in the resource group or use the default location.
          if (this.util.isNullOrEmpty(location)) {
             if (!this.util.isNullOrEmpty(resourceGroup)) {
+                // Check if Container App exists in the resource group provided and get the location from the Container App Environment linked to it
                 let doesContainerAppExist = await this.appHelper.doesContainerAppExist(this.containerAppName, resourceGroup);
+                // Check if Container App Environment is provided and exits in the resource group provided
                 let doesContainerAppEnvironmentExist = !this.util.isNullOrEmpty(containerAppEnvironment) ? await this.appHelper.doesContainerAppEnvironmentExist(containerAppEnvironment, resourceGroup) : null;
                 if (doesContainerAppExist) {
+                    // Get the name of the Container App Environment linked to the Container App
                     var environmentName = await this.appHelper.getExistingContainerAppEnvironmentName(this.containerAppName, resourceGroup);
                     // Check if environment exists in the resource group provided and get the location
                     var doesContainerAppEnvironmentExistInResourceGroup = !this.util.isNullOrEmpty(environmentName) ? await this.appHelper.doesContainerAppEnvironmentExist(environmentName, resourceGroup) : null;
                     if (doesContainerAppEnvironmentExistInResourceGroup){
+                        // Get the location of the Container App Environment linked to the Container App
                         location = await this.appHelper.getExistingContainerAppEnvironmentLocation(environmentName, resourceGroup);
                         return location;
                     }
                 }
+
+                // Check if Container App Environment exists in the resource group provided and get the location
                 if (doesContainerAppEnvironmentExist) {
                     location = await this.appHelper.getExistingContainerAppEnvironmentLocation(containerAppEnvironment, resourceGroup);
                     return location;
                 }
             }
 
+            // Get the default location for if the Container App or Container App Environment was not found in the resource group provided.
             location = await this.appHelper.getDefaultContainerAppLocation();
         }
 
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -236,10 +236,11 @@ export class azurecontainerapps {
 
         // If no location was provided, attempt to discover the location of the existing Container App Environment linked to the Container App
         // or Container App Environment provided in the resource group or use the default location.
-        let resourceGroup: string = this.toolHelper.getInput('resourceGroup', false);
-
         // Get the resource group if it was provided, or generate it from the Container App name
-        !this.util.isNullOrEmpty(resourceGroup) ? resourceGroup : `${this.containerAppName}-rg`;
+        let resourceGroup: string = this.toolHelper.getInput('resourceGroup', false);
+        if (this.util.isNullOrEmpty(resourceGroup)){
+            resourceGroup = `${this.containerAppName}-rg`;
+        }
 
         // Check if Container App exists in the resource group provided and get the location from the Container App Environment linked to it
         this.containerAppExists = await this.appHelper.doesContainerAppExist(this.containerAppName, resourceGroup);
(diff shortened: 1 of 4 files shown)


cormacpayne (approved):

src/ContainerAppHelper.ts
```diff
@@ -270,6 +270,38 @@ export class ContainerAppHelper {
...
+    /**
+     * Gets the location of an existing Container App Environment
+     * @param environmentName - the name of the Container App Environment
+     * @param resourceGroup - the resource group that the Container App Environment is found in
+    */
+    public async getExistingContainerAppEnvironmentLocation(environmentName: string, resourceGroup: string) {
+        try {
+            let command = `az containerapp env show -g ${resourceGroup} --query location -n ${environmentName}`;
+            let executionResult = await util.execute(command);
+            return executionResult.exitCode === 0 ? executionResult.stdout.toLowerCase().replace(/["() ]/g, "").trim() : null;
```
daniv-msft, line 282: If I understood properly, this logic is converting the region names from user friendly "North Central US (Stage)" to "northcentralusstage"? Not a blocker for this PR as I believe this will work in practice, but I believe we should be able to pass the user friendly regions as arguments as well, as long as they are between quotes.

azurecontainerapps.ts
```diff
@@ -232,9 +232,36 @@ export class azurecontainerapps {
...
+        // or Container App Environment provided in the resource group or use the default location.
         if (this.util.isNullOrEmpty(location)) {
+            if (!this.util.isNullOrEmpty(resourceGroup)) {
+                // Check if Container App exists in the resource group provided and get the location from the Container App Environment linked to it
+                let doesContainerAppExist = await this.appHelper.doesContainerAppExist(this.containerAppName, resourceGroup);
+                if (doesContainerAppExist) {
+                    // Get the name of the Container App Environment linked to the Container App
+                    var environmentName = await this.appHelper.getExistingContainerAppEnvironmentName(this.containerAppName, resourceGroup);
+                    // Check if environment exists in the resource group provided and get the location
+                    var doesContainerAppEnvironmentExistInResourceGroup = !this.util.isNullOrEmpty(environmentName) ? await this.appHelper.doesContainerAppEnvironmentExist(environmentName, resourceGroup) : null;
```
daniv-msft, line 248: Should the ": null;" be replaced by ": false;" ? As `doesContainerAppEnvironmentExistInResourceGroup` is a boolean, it seems weird to assign `null` to it.

azurecontainerapps.ts
```diff
@@ -232,9 +232,36 @@ export class azurecontainerapps {
...
+                    if (doesContainerAppEnvironmentExistInResourceGroup) {
+                        // Get the location of the Container App Environment linked to the Container App
+                        location = await this.appHelper.getExistingContainerAppEnvironmentLocation(environmentName, resourceGroup);
+                        return location;
+                    }
+                }
+
+                // Check if Container App Environment is provided and exits in the resource group provided and get the location
+                let doesContainerAppEnvironmentExist = !this.util.isNullOrEmpty(containerAppEnvironment) ? await this.appHelper.doesContainerAppEnvironmentExist(containerAppEnvironment, resourceGroup) : null;
```
daniv-msft, line 257: Could you please confirm that at this point we know that containerAppEnvironment is set? I can't find the condition validating it.

azurecontainerapps.ts
```diff
@@ -232,9 +232,36 @@ export class azurecontainerapps {
     private static async getLocation(): Promise<string> {
         // Set deployment location, if provided
         let location: string = this.toolHelper.getInput('location', false);
+        let resourceGroup: string = this.toolHelper.getInput('resourceGroup', false);
+        let containerAppEnvironment: string = this.toolHelper.getInput('containerAppEnvironment', false);
 
-        // If no location was provided, use the default location for the Container App service
+        // If no location was provided, attempt to discover the location of the existing Container App Environment linked to the Container App
+        // or Container App Environment provided in the resource group or use the default location.
         if (this.util.isNullOrEmpty(location)) {
```
daniv-msft, line 240: I might have missed it, but if `location` is set, when do we use that value?
Also, please consider inverting your conditions and returning early, as else the code becomes nested and hard to follow. So for example, you could have:
```
if (!this.util.isNullOrEmpty(location)) {
    return location;
}

if (<next condition>) {
    executeXYZ
    return XYZ
}
```

azurecontainerapps.ts
snehapar9: Makes sense, updated

src/ContainerAppHelper.ts
snehapar9: Just so I'm clear , if it's provided as an argument from the customer I think it will be `northcentralusstage` but if we extracted from the `show` command, afaik we need to convert it to the user friendly format.
Did that answer this question?

azurecontainerapps.ts
snehapar9: This line of code `let containerAppEnvironment: string = this.toolHelper.getInput('containerAppEnvironment', false);` is executed before this code on line 257. So, we are checking if the `containerAppEnvironment` argument was provided by the customer and exists in the resource group. 

If it does not, we use the default location.

azurecontainerapps.ts
snehapar9: Ok, makes sense. Updated this.

azurecontainerapps.ts
```diff
@@ -232,13 +232,43 @@ export class azurecontainerapps {
     private static async getLocation(): Promise<string> {
         // Set deployment location, if provided
         let location: string = this.toolHelper.getInput('location', false);
+        let resourceGroup: string = this.toolHelper.getInput('resourceGroup', false);
+        let containerAppEnvironment: string = this.toolHelper.getInput('containerAppEnvironment', false);
```
cormacpayne, line 236: nit: can we move these lines to after the below if-statement since they're not used until after it?

azurecontainerapps.ts
```diff
@@ -232,13 +232,43 @@ export class azurecontainerapps {
...
+        if (!this.util.isNullOrEmpty(location)) {
+            return location;
+        }
 
-        // If no location was provided, use the default location for the Container App service
-        if (this.util.isNullOrEmpty(location)) {
-            location = await this.appHelper.getDefaultContainerAppLocation();
+        // If no location was provided, attempt to discover the location of the existing Container App Environment linked to the Container App
+        // or Container App Environment provided in the resource group or use the default location.
+
```
cormacpayne, line 244: nit: can we remove this empty line?

src/ContainerAppHelper.ts
daniv-msft: Honestly I don't think we need to worry too much about this for now, as in practice the logic here should work I expect. My comment was just that potentially we might not even need to convert?
Whatever the format we get to identify the region ("northcentralusstage" or "North Central US (Stage)"), I expect that we should be able to pass this without conversion to the other commands as long as we put the region name between quotes. Again, no need to look into this more for this PR: I think the current approach is OK as well.

azurecontainerapps.ts
```diff
@@ -232,13 +232,43 @@ export class azurecontainerapps {
...
+        if (!this.util.isNullOrEmpty(resourceGroup)) {
+            // Check if Container App exists in the resource group provided and get the location from the Container App Environment linked to it
+            let doesContainerAppExist = await this.appHelper.doesContainerAppExist(this.containerAppName, resourceGroup);
+            if (doesContainerAppExist) {
+                // Get the name of the Container App Environment linked to the Container App
+                var environmentName = await this.appHelper.getExistingContainerAppEnvironmentName(this.containerAppName, resourceGroup);
+                // Check if environment exists in the resource group provided and get the location
```
cormacpayne, line 251: nit: can we put an empty line before this line with a comment?

src/ContainerAppHelper.ts
```diff
@@ -270,6 +270,38 @@ export class ContainerAppHelper {
...
+    /**
+     * Gets the environment Id of an existing Container App
+     * @param containerAppName - the name of the Container App
+     * @param resourceGroup - the resource group that the Container App is found in
+    */
+    public async getExistingContainerAppEnvironmentName(containerAppName: string, resourceGroup: string) {
+        try {
+            let command = `az containerapp show -n ${containerAppName} -g ${resourceGroup} --query properties.environmentId`;
+            let executionResult = await util.execute(command);
+            return executionResult.exitCode === 0 ? executionResult.stdout.split("/").pop() : null;
```
cormacpayne, line 298: We should verify that the environment ID returned doesn't end in `/` or else this will return an empty string.

src/ContainerAppHelper.ts
```diff
@@ -270,6 +270,38 @@ export class ContainerAppHelper {
...
+            let executionResult = await util.execute(command);
+            return executionResult.exitCode === 0 ? executionResult.stdout.toLowerCase().replace(/["() ]/g, "").trim() : null;
+        } catch (err) {
+            toolHelper.writeInfo(err.message);
+            return null;
+        }
+    }
+
+    /**
+     * Gets the environment Id of an existing Container App
```
cormacpayne, line 290: nit: _Gets the environment **name** of an existing Container App_

azurecontainerapps.ts
```diff
@@ -232,13 +232,43 @@ export class azurecontainerapps {
...
 
-        // If no location was provided, use the default location for the Container App service
-        if (this.util.isNullOrEmpty(location)) {
-            location = await this.appHelper.getDefaultContainerAppLocation();
+        // If no location was provided, attempt to discover the location of the existing Container App Environment linked to the Container App
+        // or Container App Environment provided in the resource group or use the default location.
+
+        if (!this.util.isNullOrEmpty(resourceGroup)) {
+            // Check if Container App exists in the resource group provided and get the location from the Container App Environment linked to it
+            let doesContainerAppExist = await this.appHelper.doesContainerAppExist(this.containerAppName, resourceGroup);
```
cormacpayne, line 247: There's an existing global variable `containerAppExists` that we should set the value of here. Anywhere else in the code that this variable is used we should check to see if its value is already set to avoid making an additional Azure CLI call to verify if the Container App exists or not.

cormacpayne (approved):

azurecontainerapps.ts

azurecontainerapps.ts
snehapar9: Updated.

azurecontainerapps.ts

src/ContainerAppHelper.ts
snehapar9: Updated.

src/ContainerAppHelper.ts
snehapar9: Updated.

azurecontainerapps.ts

azurecontainerapps.ts
```diff
@@ -201,9 +201,6 @@ export class azurecontainerapps {
         // Get the resource group to deploy to if it was provided, or generate it from the Container App name
         this.resourceGroup = await this.getOrCreateResourceGroup(this.containerAppName, this.location);
 
-        // Determine if the Container App currently exists
-        this.containerAppExists = await this.appHelper.doesContainerAppExist(this.containerAppName, this.resourceGroup);
-
         // If the Container App doesn't exist, get/create the Container App Environment to use for the Container App
         if (!this.containerAppExists) {
```
daniv-msft, line 205: I'm a bit confused about removing the `this.containerAppExists` code line 204-206. Isn't it going to have an impact on this condition here?

azurecontainerapps.ts
```diff
@@ -233,12 +230,46 @@ export class azurecontainerapps {
...
+        if (!this.util.isNullOrEmpty(location)) {
+            return location;
+        }
+
+        // If no location was provided, attempt to discover the location of the existing Container App Environment linked to the Container App
+        // or Container App Environment provided in the resource group or use the default location.
+        let resourceGroup: string = this.toolHelper.getInput('resourceGroup', false);
+
+        // Get the resource group if it was provided, or generate it from the Container App name
+        !this.util.isNullOrEmpty(resourceGroup) ? resourceGroup : `${this.containerAppName}-rg`;
```
daniv-msft, line 242: This line does not assign the values anywhere. I believe that the `this.toolHelper.getInput` method has an argument to give a default value. Should we use that argument to give `${this.containerAppName}-rg`?

azurecontainerapps.ts
cormacpayne: Yeah we should keep this line but only execute it if the value of `this.containerAppExists` hasn't been set previously (we may need to make the variable a nullable boolean so we can differentiate between the default `false` value and no value set yet).

azurecontainerapps.ts
cormacpayne: +1

azurecontainerapps.ts
snehapar9: I moved this condition to [this line](https://github.com/Azure/container-apps-deploy-action/blob/snehapar/Remove-hardcoded-location/azurecontainerapps.ts#L245) to make sure are not making this call twice. 

So, it should not impact the rest of the flow if this line is removed from here.

azurecontainerapps.ts
daniv-msft, line 242: I'm a bit confused about generating a new resource group name. In the context of this method, we're looking for the location. In this context, if the resource group isn't available, then we know already that we should just return the default location, right? I'm not clear in which case does having a generated resource group name help us.

azurecontainerapps.ts
cormacpayne: We still need this line because if the user provides a value for `location`, then the call to `getLocation()` returns before it hits the part where it sets `this.containerAppExists` -- we just need to check if we've previously set a value for this variable or not because we don't want to call `doesContainerAppExist` multiple times.

daniv-msft: @snehapar9 I don't think we can rely on a guarantee on the order in which `setupResources` and `getLocation` will be called, as even if it is true today this could easily change (and no one would expect these to have a consequnce on each other).
So in this case, we could reach the line ` if (!this.containerAppExists) {` here and this would have an impact on the flow.

Just to be sure I got it, why did we need it to change things here? Could we handle this case in getLocation and check for the value there if it isn't already set?

azurecontainerapps.ts
snehapar9: Addressed this in my latest commit - https://github.com/Azure/container-apps-deploy-action/pull/76/commits/4faa46488fafdd70f3fbfed2a361df52a3c1c0df

azurecontainerapps.ts
snehapar9: Synced offline. Reverted this change.

daniv-msft (approved):

PR #81, opened 2023-11-28
Fixing ACR credential issues when up command was used

harryli0108:
Fixing the issue where ACR authentication fails when using up command by swapping `up` with `containerapp update` and added `ingress update`

### DIFF
--- a/.github/workflows/run-validation-oidc.yml
+++ b/.github/workflows/run-validation-oidc.yml
@@ -297,7 +297,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       id-token: write # This is required for requesting the OIDC JWT Token
-    timeout-minutes: 15
+    timeout-minutes: 25
 
     env:
       TEST_CONTAINER_APP_NAME: 'gh-ca-is-ne-${{ github.run_id }}'
@@ -607,7 +607,7 @@ jobs:
       - name: Update Container App with existing image
         if: ${{ always() }}
         shell: bash
-        run: az containerapp update -n ${{ env.TEST_CONTAINER_APP_NAME }} -g ${{ vars.TEST_RESOURCE_GROUP_NAME }} -i mcr.microsoft.com/azuredocs/containerapps-helloworld:latest
+        run: az containerapp update -n ${{ env.TEST_EXISTING_CONTAINER_APP }} -g ${{ vars.TEST_RESOURCE_GROUP_NAME }} -i mcr.microsoft.com/azuredocs/containerapps-helloworld:latest
 
   update-using-image:
 
--- a/.github/workflows/run-validation-oidc.yml
+++ b/.github/workflows/run-validation-oidc.yml
@@ -110,7 +110,7 @@ jobs:
       - name: Delete created Azure Container App
         if: ${{ always() }}
         shell: bash
-        run: az containerapp delete -n ${{ env.TEST_CONTAINER_APP_NAME }} -g ${{ vars.TEST_RESOURCE_GROUP_NAME }} -y
+        run: az containerapp delete -n ${{ env.TEST_CONTAINER_APP_NAME }} -g ${{ vars.TEST_RESOURCE_GROUP_NO_ACR_NAME }} -y
 
 
   create-using-found-dockerfile:
@@ -600,14 +600,14 @@ jobs:
         uses: ./
         with:
           appSourcePath: '${{ github.workspace }}/oryx/tests/SampleApps/DotNetCore/NetCore6PreviewWebApp'
-          containerAppName: ${{ env.TEST_EXISTING_CONTAINER_APP }}
+          containerAppName: ${{ vars.TEST_EXISTING_CONTAINER_APP }}
           resourceGroup: ${{ vars.TEST_RESOURCE_GROUP_NO_ACR_NAME }}
           disableTelemetry: ${{ vars.TEST_DISABLE_TELEMETRY }}
 
       - name: Update Container App with existing image
         if: ${{ always() }}
         shell: bash
-        run: az containerapp update -n ${{ env.TEST_EXISTING_CONTAINER_APP }} -g ${{ vars.TEST_RESOURCE_GROUP_NAME }} -i mcr.microsoft.com/azuredocs/containerapps-helloworld:latest
+        run: az containerapp update -n ${{ vars.TEST_EXISTING_CONTAINER_APP }} -g ${{ vars.TEST_RESOURCE_GROUP_NO_ACR_NAME }} -i mcr.microsoft.com/azuredocs/containerapps-helloworld:latest
 
   update-using-image:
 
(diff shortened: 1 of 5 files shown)


harryli0108:
Please don't merge before team member approved this change.

azurecontainerapps.ts
```diff
@@ -608,8 +608,8 @@ export class azurecontainerapps {
         } else if (this.shouldCreateOrUpdateContainerAppWithUp) {
             await this.appHelper.createOrUpdateContainerAppWithUp(this.containerAppName, this.resourceGroup, this.commandLineArgs);
         } else {
-            // Update the Container App using the 'up' command
-            await this.appHelper.updateContainerAppWithUp(this.containerAppName, this.resourceGroup, this.commandLineArgs, this.ingress, this.targetPort);
+            // Update the Container App using the 'containerapp update' and 'ingress update' commands
+            await this.appHelper.updateContainerAppWithUpdateAndIngress(this.containerAppName, this.resourceGroup, this.commandLineArgs, this.ingress, this.targetPort);
```
cormacpayne, line 612: I would be in favor of us having two separate calls here: one that updates the Container App and another that updates the ingress for the Container App

```typescript
// Update the Container App using the 'containerapp update' command
await this.appHelper.updateContainerAppWithUpdate(this.containerAppName, this.resourceGroup, this.commandLineArgs);

// Update the ingress and target port on the Container App using the 'ingress update' command
await this.appHelper.updateContainerAppIngress(this.containerAppName, this.resourceGroup, this.ingress, this.targetPort);
```

This will allow the functions to not be overloaded with functionality and be reusable in other scenarios we may want to use them in.

cormacpayne (changes requested):

azurecontainerapps.ts
harryli0108: makes sense! changing this in the next commit

azurecontainerapps.ts
```diff
@@ -608,8 +608,9 @@ export class azurecontainerapps {
         } else if (this.shouldCreateOrUpdateContainerAppWithUp) {
             await this.appHelper.createOrUpdateContainerAppWithUp(this.containerAppName, this.resourceGroup, this.commandLineArgs);
         } else {
-            // Update the Container App using the 'up' command
-            await this.appHelper.updateContainerAppWithUp(this.containerAppName, this.resourceGroup, this.commandLineArgs, this.ingress, this.targetPort);
+            // Update the Container App using the 'containerapp update' and 'ingress update' commands
+            await this.appHelper.updateContainerApp(this.containerAppName, this.resourceGroup, this.commandLineArgs)
```
snehapar9, line 612: Synced offline, since we don't use the `up` command to update the container app anymore, we should change the name of the global variable `this.shouldUseUpdateCommand` to clearly indicate when the `update ingress` command needs to be executed.

src/ContainerAppHelper.ts
```diff
@@ -147,6 +147,37 @@ export class ContainerAppHelper {
         }
     }
 
+    /**
+     * Update container app with update and ingress update to avoid failure of acr authentication.
+     * @param containerAppName - the name of the existing Container App
+     * @param resourceGroup - the resource group that the existing Container App is found in
+     * @param optionalCmdArgs - a set of optional command line arguments
```
snehapar9, line 154: Can we remove `optionalCmdArgs` on line 154? I'm not seeing it in the method `updateContainerAppIngress` below?

azurecontainerapps.ts
harryli0108: sounds good! updated in the latest commit

src/ContainerAppHelper.ts
harryli0108: makes sense. fixed in the latest commit.

.github/workflows/run-validation-oidc.yml
```diff
@@ -592,22 +592,22 @@ jobs:
...
+          client-id: ${{ secrets.OIDC_TEST_CLIENT_ID }}
+          tenant-id: ${{ secrets.OIDC_TEST_TENANT_ID }}
+          subscription-id: ${{ secrets.OIDC_TEST_SUBSCRIPTION_ID }}
 
       - name: Execute Azure Container Apps Build and Deploy Action
         uses: ./
         with:
           appSourcePath: '${{ github.workspace }}/oryx/tests/SampleApps/DotNetCore/NetCore6PreviewWebApp'
-          containerAppName: ${{ env.TEST_EXISTING_CONTAINER_APP }}
+          containerAppName: ${{ vars.TEST_EXISTING_CONTAINER_APP }}
```
cormacpayne, line 603: What's the reason for this change? It looks like other tests are still referencing `env` and not `var`

src/ContainerAppHelper.ts
```diff
@@ -147,6 +147,36 @@ export class ContainerAppHelper {
         }
     }
 
+    /**
+     * Update container app with update and ingress update to avoid failure of acr authentication.
```
cormacpayne, line 151: nit: it would be great to re-word this comment:

_Updates the ingress on an existing Container App. Note: this helps to avoid failure during ACR authentication._

azurecontainerapps.ts
```diff
@@ -502,13 +503,14 @@ export class azurecontainerapps {
...
-        this.shouldUseUpdateCommand = this.containerAppExists &&
+        this.noIngressUpdate = this.containerAppExists &&
             this.util.isNullOrEmpty(this.targetPort) &&
             (this.util.isNullOrEmpty(this.ingress) || this.ingress == 'disabled');
 
         // Pass the Container Registry credentials when creating a Container App or updating a Container App via the 'up' command
         if (!this.util.isNullOrEmpty(this.registryUrl) && !this.util.isNullOrEmpty(this.registryUsername) && !this.util.isNullOrEmpty(this.registryPassword) &&
-            (!this.containerAppExists || (this.containerAppExists && !this.shouldUseUpdateCommand))) {
+            (!this.containerAppExists || (this.containerAppExists && !this.noIngressUpdate))) {
+            this.adminCredentialsProvided = true;
```
cormacpayne, line 513: nit: let's leave a comment above this variable specifying what it represents and why we're setting it to `true` in this case

azurecontainerapps.ts
cormacpayne: @snehapar9 -- I'm not sure I understand your comment _since we don't use the `up` command to update the container app anymore_ because I do see a function call a few lines up `updateContainerAppWithUp` that still uses `up`. Would you mind elaborating a bit more?

cormacpayne (approved):

.github/workflows/run-validation-oidc.yml
harryli0108: This one was added because for some reason the action couldn't read the container app name (shown as empty in debug message). It then got fixed after I changed it from env to vars.

src/ContainerAppHelper.ts
harryli0108: Sounds good! I'll have a separate pr to update the comments

azurecontainerapps.ts
snehapar9: @cormacpayne This comment does not need to be addressed anymore.
@harryli0108's initial changes had `up` removed and made calls to only `update`. But since `update` does not support registry credentials, `up` was added back.

PR #68, opened 2023-10-27
Migrate latest changes from v1 to v2

cormacpayne:
Includes the changes from the following PRs:
- https://github.com/Azure/container-apps-deploy-action/pull/63

PR #82, opened 2023-11-29
support build environment variable

yilims:


README.md
```diff
@@ -171,6 +171,7 @@ For more information on the structure of the YAML configuration file, please vis
...
 | `builderStack`            | No       | The stack (OS) that should be used to build the provided application source and produce the runnable application image. You can provide a specific image tag for the stack, such as "debian-bookworm-20231107.2", or you can provide a supported stack name, such as "debian-bookworm" or "debian-bullseye", and the latest supported image tag for that stack will be used. If no stack is provided, this action will attempt to build the provided application source with each supported stack until there's a successful build. |
+| `buildEnvironmentVariables` | No     | A list of environment variable(s) used to configure the Buildpacks for image build. Space-separated values in 'key=value' format. Empty string to clear existing values. |
```
cormacpayne, line 174: We previously introduced a [`buildArguments`](https://github.com/Azure/container-apps-deploy-action/pull/68/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6R91-R97) input for this GitHub Action that would allow customers to provide custom environment variables for (1) their Dockerfile in the `docker build` flow and (2) the buildpack environment when running `pack build`.

We ended up reverting this change when it wasn't entirely clear how this would be used in the source-to-cloud flow when it was introduced in the CLI and took precedence over the local `pack build` we were running within this action.

I think in this case, it would be a good idea to bring back this `buildArguments` input to specifically handle these two cases, and later on propagate the value(s) to the CLI when the commands can handle these environment variables being provided and will automatically propagate them to the cloud build.

azurecontainerapps.ts
```diff
@@ -460,6 +483,13 @@ export class azurecontainerapps {
             environmentVariables.push(`ORYX_RUNTIME_PORT=${this.targetPort}`);
         }
 
+        // Set the user defined environment variables that should be propagated to the builder
+        if (!this.util.isNullOrEmpty(this.buildEnvironmentVariables)) {
+            this.buildEnvironmentVariables.match(/"[^"]*"|\S+/g).forEach((envVar) => {
```
cormacpayne, line 488: Let's set `/"[^"]*"|\S+/g` as a global RegEx property that we can reference instead of the direct RegEx just in case we need to make any changes to it in the future.

cormacpayne (changes requested):

README.md
cormacpayne: 1. When @daniv-msft and I chatted previously, I believe we settled on `buildArguments` since it made sense from a `docker build` and `pack build` standpoint where these arguments would be used (_i.e._, `--build-arg` for `docker build` and arguments for the "build environment" for `pack build`). Do you have any issues with making this input called `buildArguments`?

2. In the previous PR, we just used a comma to separate the values, but I don't think we had any strong input on that-- if the Azure CLI is using spaces and that's how you have it currently implemented, then I'm OK with keeping that format.

yilims: #1: Let's use `buildArguments` as you guys have made decision before.
#2: Let's use spaces to be consistent with the CLI, which may be easier for user to understand.

azurecontainerapps.ts
```diff
@@ -174,6 +180,41 @@ export class azurecontainerapps {
...
             throw Error(conflictingArgumentsMessage);
         }
+
+        // Set up the build arguments to pass to the Dockerfile or builder
+        if (!this.util.isNullOrEmpty(this.buildArguments)) {
+            // Ensure that the build arguments are in the format 'key1=value1 key2=value2'
+            const buildArguments = this.buildArguments.match(buildArgumentRegex);
+            const dockerfilePath = this.toolHelper.getInput('dockerfilePath', false);
+            const isBuildPackBuild = this.util.isNullOrEmpty(dockerfilePath);
+            let invalidBuildArgumentsMessage = `The 'buildArguments' argument must be in the format 'key1=value1 key2=value2'.`;
```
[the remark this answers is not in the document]
yilims: Yes, the RegEx handles the quotes:
const buildArgumentRegex = /"[^"]*"|\S+/g;

README.md
```diff
@@ -171,6 +171,7 @@ For more information on the structure of the YAML configuration file, please vis
...
 | `builderStack`            | No       | The stack (OS) that should be used to build the provided application source and produce the runnable application image. You can provide a specific image tag for the stack, such as "debian-bookworm-20231107.2", or you can provide a supported stack name, such as "debian-bookworm" or "debian-bullseye", and the latest supported image tag for that stack will be used. If no stack is provided, this action will attempt to build the provided application source with each supported stack until there's a successful build. |
+| `buildArguments`          | No       | A list of build arguments provided as KEY=VALUE pairings and are space-separated. If a Dockerfile has been provided or is discovered in the application source, each build argument will be passed to the `docker build` command via the `--build-arg` flag. If the Oryx++ builder is used to create a runnable application image, each build argument will be passed to the `pack build` command via the `--env` flag. |
```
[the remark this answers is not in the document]
yilims: added usage for both pack build and docker build commands.

PR #86, opened 2024-01-08
Add subscription information when to call builder

showpune:


cormacpayne:
@showpune looks like this was automatically closed when we merged `v2-main` back into `main` -- I'm going to create a separate PR with your changes in this repository against `main` so that CI will properly run

PR #88, opened 2024-01-09
Add subscription information when to call builder

cormacpayne:
Forked from https://github.com/Azure/container-apps-deploy-action/pull/86


src/ContainerAppHelper.ts
```diff
@@ -423,7 +443,7 @@ export class ContainerAppHelper {
             toolHelper.writeDebug(`Attempting to create a runnable application image with name "${imageToDeploy}" using the Oryx++ Builder "${builderImage}"`);
 
             try {
-                let command = `build ${imageToDeploy} --path ${appSourcePath} --builder ${builderImage} --env ${telemetryArg}`;
+                let command = `build ${imageToDeploy} --path ${appSourcePath} --builder ${builderImage} --env ${telemetryArg} --env ORYX_SUBSCRIPTION_ID=${subscription}`;
```
daniv-msft, line 446: Should we consider naming this differently than "ORYX_SUBSCRIPTION_ID", as this is not really related to Oryx? Is it because of the env variables prefix logic than using "ORYX_" makes things easier?

showpune: Using the BP prefix BP_SUBSCRIPTION_ID

PR #95, opened 2024-01-24
Http 400 error and env var arguments error fix

harryli0108:
1. Resolving the http 400 error when creating new container app using `az containerapp create` command
2. Fixed the error when environment variables are appended to `az containerapp up` command

### DIFF
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -344,10 +344,9 @@ export class azurecontainerapps {
 
         // See if we can reuse an existing Container App environment found in the resource group
         if (this.util.isNullOrEmpty(containerAppEnvironment)) {
-            let existingContainerAppEnvironment: string = await this.appHelper.getExistingContainerAppEnvironment(resourceGroup);
+            const existingContainerAppEnvironment: string = await this.appHelper.getExistingContainerAppEnvironment(resourceGroup);
             if (!this.util.isNullOrEmpty(existingContainerAppEnvironment)) {
                 this.toolHelper.writeInfo(`Existing Container App environment found in resource group: ${existingContainerAppEnvironment}`);
-                existingContainerAppEnvironment = existingContainerAppEnvironment.replace(/(\r\n|\n|\r)/gm, "");
                 return existingContainerAppEnvironment
             }
         }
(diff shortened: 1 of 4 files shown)


src/ContainerAppHelper.ts
```diff
@@ -293,7 +293,7 @@ export class ContainerAppHelper {
         try {
             let command = `az containerapp env list -g ${resourceGroup} --query "[0].name"`
             let executionResult = await util.execute(command);
-            return executionResult.exitCode === 0 ? executionResult.stdout : null;
+            return executionResult.exitCode === 0 ? executionResult.stdout.replace(/(\r\n|\n|\r)/gm, "") : null;
```
daniv-msft, line 296: Nit: should we consider making this at the util.execute level? It feels like this could be a problem in all cases potentially?

azurecontainerapps.ts
```diff
@@ -606,12 +606,12 @@ export class azurecontainerapps {
         }
 
         const environmentVariables: string = this.toolHelper.getInput('environmentVariables', false);
-
+        const isCappUpdateCommandUsed: boolean = this.noIngressUpdate || (!this.noIngressUpdate && !this.adminCredentialsProvided)
```
daniv-msft, line 609: Could you please share more info about that part of the change? Why do we need this change?

src/ContainerAppHelper.ts
snehapar9: +1. We can probably add this `replace` method to stdout [here](https://github.com/Azure/container-apps-deploy-action/blob/main/src/GithubActionsToolHelper.ts#L52) so that way we know the standard output does not have `/r` or `/n` as such

azurecontainerapps.ts
harryli0108: Yes, this check was added because we had two different ways to add env vars when running containerapp create, up, and update. For `update`, the option name was `---replace-env-vars`. And for any other commands we used, it was `--env-vars`.

Due to one of the recent changes we made with ingress update (Issue: https://github.com/microsoft/azure-container-apps/issues/821 PR: #81), the wrong argument was appended to `update` command, which resulted in some user running into issue when they have ingress and env variables set.

And as this check was named, it checks if the situation is going to use `update` command. If it is, then `--replace-env-vars` option will be appended.

src/ContainerAppHelper.ts
harryli0108: Makes sense! Moving it in next commit

azurecontainerapps.ts
daniv-msft: OK, thanks for sharing more info!

Issue #101, opened 2024-02-13
ERROR: unrecognized arguments: --replace-env-vars


moderj:
Hey I'm noticing an issue regarding the `environmentVariables` flag

This is the error it throws when running github action
`ERROR: unrecognized arguments: --replace-env-vars LOG_LEVEL=info DB_URL=secretref:*** DB_NAME=secretref:***`
`Error: The process '/usr/bin/az' failed with exit code 2
Error: The process '/usr/bin/az' failed with exit code 2`

This is the stage:
![Screenshot 2024-02-13 102020](https://github.com/Azure/container-apps-deploy-action/assets/118797659/89bfa570-75f9-4ea3-808f-887a4a4d0c01)

amitkma:
We are facing the same issue. Is there any update on this?

PR #106, opened 2024-04-25
Fix unrecognized argument `replace-env-vars` when creating container

amitkma:
This PR fixes issue #101. When a container is supposed to be created, `isCappUpdateCommandUsed` variable is  set to True incorrectly.


amitkma:
@microsoft-github-policy-service agree company="DotMyStyle"

PR #107, opened 2024-05-14
Use create/update to create or update container apps in source to cloud scenarios

snehapar9:
This change ensures container apps are created/updated using `create`/`update` command instead of `up` in source to cloud scenarios.

### DIFF
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -18,7 +18,7 @@ export class azurecontainerapps {
             this.validateSupportedScenarioArguments();
 
             // Set up the Azure CLI to be used for this task
-            await this.setupAzureCli();
+        //    await this.setupAzureCli();
 
             // Set up the resources required to deploy a Container App
             await this.setupResources();
@@ -207,7 +207,7 @@ export class azurecontainerapps {
      */
     private static async setupAzureCli() {
         // Set the Azure CLI to install missing extensions
-        await this.util.installAzureCliExtension();
+     //   await this.util.installAzureCliExtension();
     }
 
     /**
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -106,7 +106,6 @@ export class azurecontainerapps {
     private static buildArguments: string;
     private static noIngressUpdate: boolean;
     private static useInternalRegistry: boolean;
-    // private static shouldCreateOrUpdateContainerAppWithUp: boolean;
 
     /**
      * Initializes the helpers used by this task.
--- a/azurecontainerapps.ts
+++ b/azurecontainerapps.ts
@@ -39,7 +39,7 @@ export class azurecontainerapps {
             // Set up property to trigger cloud build with 'up' command
             // Do not run up for source to cloud build scenario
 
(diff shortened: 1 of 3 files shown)


azurecontainerapps.ts
```diff
@@ -109,7 +106,7 @@ export class azurecontainerapps {
     private static buildArguments: string;
     private static noIngressUpdate: boolean;
     private static useInternalRegistry: boolean;
-    private static shouldCreateOrUpdateContainerAppWithUp: boolean;
+    // private static shouldCreateOrUpdateContainerAppWithUp: boolean;
```
pauld-msft, line 109: remove

azurecontainerapps.ts
```diff
@@ -653,16 +646,14 @@ export class azurecontainerapps {
...
                 await this.appHelper.updateContainerAppRegistryDetails(this.containerAppName, this.resourceGroup, this.registryUrl, this.registryUsername, this.registryPassword);
             }
 
             // Update the Container App using the 'update' command
             await this.appHelper.updateContainerApp(this.containerAppName, this.resourceGroup, this.commandLineArgs);
-        } else if (this.shouldCreateOrUpdateContainerAppWithUp) {
-            await this.appHelper.createOrUpdateContainerAppWithUp(this.containerAppName, this.resourceGroup, this.commandLineArgs);
         } else if (this.adminCredentialsProvided && !this.noIngressUpdate) {
             // Update the Container App with `up` command when admin credentials are provided and ingress is manually provided.
             await this.appHelper.updateContainerAppWithUp(this.containerAppName, this.resourceGroup, this.commandLineArgs, this.ingress, this.targetPort);
```
pauld-msft, line 659: we still appear to be updating with the Up command here

azurecontainerapps.ts
```diff
@@ -621,11 +618,9 @@ export class azurecontainerapps {
         // Ensure '-i' argument and '--source' argument are not both provided
         if (!this.util.isNullOrEmpty(this.imageToDeploy)) {
             this.commandLineArgs.push(`-i ${this.imageToDeploy}`);
-        } else if (this.shouldCreateOrUpdateContainerAppWithUp) {
+        } else if (!this.util.isNullOrEmpty(this.appSourcePath) && this.useInternalRegistry) {
             this.commandLineArgs.push(`--source ${this.appSourcePath}`);
-            this.commandLineArgs.push(`-l ${this.location}`);
```
pauld-msft, line 626: is location arg not allowed / not working for `container app update`?

snehapar9: nope `update` does not support location. It derives it from env

azurecontainerapps.ts
snehapar9: Yes, we still need these. This change just updates container app for only source to cloud scenarios. There are other scenarios that still require up and is already being handled

pauld-msft (approved):

azurecontainerapps.ts
snehapar9: removed

harryli0108 (changes requested):
LGTM :shipit:

harryli0108 (approved):

PR #110, opened 2024-07-01
Support custom container app environment resource group

benchiverton:
## Background
Within my cloud infrastructure, I have a shared container app environment. When creating a new environment, I create an instance-level resource group to deploy my container apps to. I do this so that I can logically group my resources and avoid creating multiple/destroying app environments (which takes a while!).

However, this action assumes that the container app environment + container app are in the same rg. This PR aims to add functionality to support scenarios where the container app environment + container app are in the same rg.

## Proposed solution
Add a new parameter `containerAppEnvironmentResourceGroup`, which when set will look for / create the container app environment in the supplied rg. If it's not set, resource groups will be created in the same way.

#### Questions / known issues
* I've got no idea how to regenerate the index.js file, if someone could help me out that would be massively appreciated!


action.yml
```diff
@@ -72,6 +72,11 @@ inputs:
       environment in the resource group of the Container App will be used, otherwise, an environment will be created in
       the format <container-app-name>-env'
     required: false
+  containerAppEnvironmentResourceGroup:
+    description: |
+      'The name of the Resource Group that contains the Azure Container App environment to use with the application.
+      If not provided, it will fall back to the valie of resourceGroup, which defaults to <container-app-name>-env'
```
TonyArra, line 78:
```suggestion
      If not provided, it will fall back to the value of resourceGroup, which defaults to <container-app-name>-env'
```

README.md
```diff
@@ -169,6 +169,7 @@ For more information on the structure of the YAML configuration file, please vis
 | `containerAppName`        | No       | The name of the Container App that will be created or updated. If not provided, this value will be `github-action-container-app-<github-run-id>-<github-run-attempt>`. |
 | `resourceGroup`           | No       | The existing resource group that the Azure Container App will be created in. If not provided, this value will be `<container-app-name>-rg` and its existence will first be checked before attempting to create it. |
 | `containerAppEnvironment` | No       | The name of the Container App environment to use with the application. If not provided, an existing environment in the resource group of the Container App will be used, otherwise, an environment will be created in the formation `<container-app-name>-env`. |
+| `containerAppEnvironmentResourceGroup` | No       | The name of the Resource Group that contains the Azure Container App environment to use with the application. If not provided, it will fall back to the valie of resourceGroup, which defaults to `<container-app-name>-env`. |
```
TonyArra, line 172:
```suggestion
| `containerAppEnvironmentResourceGroup` | No       | The name of the Resource Group that contains the Azure Container App environment to use with the application. If not provided, it will fall back to the value of resourceGroup, which defaults to `<container-app-name>-env`. |
```

PR #133, opened 2025-09-19
Fix bug on building command

mjy9088:


mjy9088:
@microsoft-github-policy-service agree

mjy9088:
Connection strings on Azure services, such as Application Insights and Azure Storage contain semicolons.

`environmentVariables` couldn't contain semicolons before this fix was applied.

